PR #2, opened 2017-07-19
run provider generator to sync changes from core

durandom:
ran the generator again with

```bash
bundle exec rails g provider <name> --path ../
```

these are the changes.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -10,3 +10,4 @@ end
 #end
 
 Dir[Rails.root.join("spec/shared/**/*.rb")].each { |f| require f }
+Dir[ManageIQ::Providers::Nuage::Engine.root.join("spec/support/**/*.rb")].each { |f| require f }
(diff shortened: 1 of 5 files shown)


spec/spec_helper.rb
```diff
@@ -3,10 +3,10 @@
...
 
-# Uncomment in case you use vcr cassettes
-#VCR.configure do |config|
-#  config.ignore_hosts 'codeclimate.com' if ENV['CI']
-#  config.cassette_library_dir = File.join(ManageIQ::Providers::Nuage::Engine.root, 'spec/vcr_cassettes')
-#end
+VCR.configure do |config|
+  config.ignore_hosts 'codeclimate.com' if ENV['CI']
+  config.cassette_library_dir = File.join(ManageIQ::Providers::Nuage::Engine.root, 'spec/vcr_cassettes')
+end
```
Fryguy: This provider does not use VCR casettes.

durandom:
@Fryguy updated

PR #3, opened 2017-07-21
Add the missing excon gem

gberginc:
Without this gem I was getting error about the missing Excon constant while trying to log into Nuage provider. This patch adds the same version as the one currently used in OpenStack provider.


juliancheal:
LTGM 👍

juliancheal:
@miq-bot assign @bronaghs

PR #4, opened 2017-08-18
Fix a problem when no policy groups exist

gberginc:
In case no policy groups are retrieved from the Nuage provider, `nil` is returned causing a problem processing the collection. This patch makes a simple check to skip processing should the list be nil.

@pdellaert Is empty list of policy groups an indication of some other error that should not be handled just by skipping them in the Nuage parser?

@miq-bot assign @juliancheal 
@miq-bot add_label bug


pdellaert:
@gberginc An empty policy groups list is possible and not an error situation. Thanks for fixing this!

PR #6, opened 2017-08-22
Introduce VCR tests for Nuage network manager

gberginc:
The Nuage network manager did previously not provide anky kind of automated tests causing refactoring and implementation of additional features rather risky (and time consuming). This patch introduces the initial spec that uses VCR recordings against a testing VSD environment. The spec currently only verifies the table counts as the refresh parser does not populate much data into the database.

The patch obfuscates the actual Nuage backend (hostname, username/password).

@miq-bot assign @juliancheal 
@miq-bot add_label refactoring

### DIFF
--- a/spec/factories/ext_management_system.rb
+++ b/spec/factories/ext_management_system.rb
@@ -1,0 +1,24 @@
+FactoryGirl.define do
+  factory :ems_nuage_with_vcr_authentication, :parent => :ems_nuage_network do
+    zone do
+      _guid, _server, zone = EvmSpecHelper.create_guid_miq_server_zone
+      zone
+    end
+
+    after(:build) do |ems|
+      ems.hostname = Rails.application.secrets.nuage_network.try(:[], 'host') || 'nuagenetworkhost'
+    end
+
+    after(:create) do |ems|
+      userid   = Rails.application.secrets.nuage_network.try(:[], 'userid') || 'NUAGE_USER_ID'
+      password = Rails.application.secrets.nuage_network.try(:[], 'password') || 'NUAGE_PASSWORD'
+
+      cred = {
+        :userid   => userid,
+        :password => password
+      }
+
+      ems.authentications << FactoryGirl.create(:authentication, cred)
+    end
+  end
+end
--- a/spec/models/manageiq/providers/nuage/network_manager/vcr_specs/refresher_spec.rb
+++ b/spec/models/manageiq/providers/nuage/network_manager/vcr_specs/refresher_spec.rb
@@ -1,0 +1,48 @@
+describe ManageIQ::Providers::Nuage::NetworkManager::Refresher do
+  before(:each) do
+    @ems = FactoryGirl.create(:ems_nuage_with_vcr_authentication, :port => 8443, :api_version => "v5_0", :security_protocol => "ssl-with-validation")
+  end
+
+  before(:each) do
+    settings                          = OpenStruct.new
+    settings.inventory_object_refresh = false
+
+    allow(Settings.ems_refresh).to receive(:nuage_network).and_return(settings)
+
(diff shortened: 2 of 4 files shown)


spec/factories/ext_management_system.rb
```diff
@@ -0,0 +1,34 @@
...
+    after(:build) do |ems|
+      ems.hostname = Rails.application.secrets.nuage_network.try(:[], 'host') || 'nuagenetworkhost'
+    end
+
+    after(:create) do |ems|
+      userid   = Rails.application.secrets.nuage_network.try(:[], 'userid') || 'NUAGE_USER_ID'
+      password = Rails.application.secrets.nuage_network.try(:[], 'password') || 'NUAGE_PASSWORD'
+
+      # Ensure that VCR will obfuscate the basic auth
+      VCR.configure do |c|
```
gberginc: With this factory girl I followed the example by the Amazon provider ([here](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/spec/factories/ext_management_system.rb#L9)). Since the actual Nuage backend and it's credentials should not be shared publicly, I decided to use a similar approach to the one in VMware cloud manager ([here](https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/spec/models/manageiq/providers/vmware/cloud_manager/refresher_spec.rb#L20-L28)).

The problem that I have is that this mixes two aspects (factory model and VCR configuration) so I wonder if it would be better to

 * move just the VCR config into the [corresponding spec](https://github.com/ManageIQ/manageiq-providers-nuage/pull/6/files#diff-3fadeabfdfdb1f0e8c5b79d3188ac70aR11) file into it's own `before(:each)`? This would again read the hostname, userid and password from `Rails.application.secrets`
 * move everything into the [corresponding spec](https://github.com/ManageIQ/manageiq-providers-nuage/pull/6/files#diff-3fadeabfdfdb1f0e8c5b79d3188ac70aR11) (as VMware does)

gberginc:
@miha-plesko Since you were involved in VMware cloud manager, could you also have a look on this?

spec/factories/ext_management_system.rb
miha-plesko: @gberginc I think configuring "VCR recording rules" inside "database object factory" makes no sense no matter what. The factory is only meant to instantiate Rails objects for us, nothing more.

As far as I can understand, the
```ruby
VCR.configure do |config|
  ...
end
```
snippet only needs to be executed once and the `NUAGE_NETWORK_AUTHORIZATION` will be replaced for every subsequent casette, right? In other words: do we really need to configure VCR before each test? I think this is not the case, therefore I suggest we move the VCR code from `spec/factories/ext_management_system.rb` to `spec/spec_helper.rb` where we merge it with current VCR setting. Would it work this way?

If configuring it in `spec/spec_helper.rb` for some reason doesn't work, then I'd put the VCR configuration code into corresponding spec as VMware does (i.e. the second bullet from your comment. THB, I don't see the difference between the two bullets that you provided - links are the same).

In both cases we will read userid and password from yaml once again, but I don't think this is a problem to worry about. That being said, I'm not sure why we even want to read it from yaml at first place - wouldn't we be better off hard-coding it? I may be missing something, but if we only use yaml for that rare occasion when VCR cassete is being recorded, then it would probably be easier to temporarily provide actual username/password into variable directly...

miha-plesko: To sum up: definetly move VCR configuration out of factory. If possible, merge with `spec_helper.rb`, if not, put directly into `refresher_spec.rb`.

gberginc:
Thanks @miha-plesko. I've decided to put the VCR obfuscation into the `refresher_spec`. In a way this feels more natural to me as it's specific to this particular test.

@juliancheal Do you think it would nevertheless be better to move that into `spec_helper` as suggested by Miha [here](https://github.com/ManageIQ/manageiq-providers-nuage/pull/6#discussion_r134747285)?

PR #9, opened 2017-09-14
Introduce Nuage event catcher

gberginc:
This commit introduces the full blown event catcher for Nuage manager based on ActiveMQ message bus. Further to default authentication, AMQP is now provided via Qpid Proton gem, along with the corresponding message handler responsible for the delivery of messages into CFME.

### DIFF
--- a/app/models/manageiq/providers/nuage/manager_mixin.rb
+++ b/app/models/manageiq/providers/nuage/manager_mixin.rb
@@ -41,18 +41,51 @@ module ManageIQ::Providers::Nuage::ManagerMixin
   def verify_credentials(auth_type = nil, options = {})
     auth_type ||= 'default'
 
+    raise MiqException::MiqInvalidCredentialsError, "Unsupported auth type: #{auth_type}" unless supports_authentication?(auth_type)
     raise MiqException::MiqHostError, "No credentials defined" if missing_credentials?(auth_type)
 
     options[:auth_type] = auth_type
+    case auth_type.to_s
+    when 'default' then verify_api_credentials(options)
+    when 'amqp'    then verify_amqp_credentials(options)
+    end
+  end
+
+  def event_monitor_options
+    @event_monitor_options ||= begin
+      url = ''
+      amqp = connection_configuration_by_role('amqp')
+      if (endpoint = amqp.try(:endpoint))
+        url = "#{endpoint.hostname}:#{endpoint.port}"
+      end
+
+      if (authentication = amqp.try(:authentication))
+        url = "#{authentication.userid}:#{authentication.password}@#{url}"
+      end
+
+      {
+        :ems                       => self,
+        :url                       => url,
+        :sasl_allow_insecure_mechs => true, # Only plain (insecure) mechanism currently supported
+      }
+    end
+  end
+
+  private
+
+  def verify_api_credentials(options = {})
     with_provider_connection(options) {}
     true
-
   rescue => err
     miq_exception = translate_exception(err)
     raise unless miq_exception
 
     _log.error("Error Class=#{err.class.name}, Message=#{err.message}")
     raise miq_exception
+  end
+
+  def verify_amqp_credentials(_options = {})
+    ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream.test_amqp_connection(event_monitor_options)
   end
 
   def auth_url(protocol, server, port, version)
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
@@ -1,0 +1,41 @@
+class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler < Qpid::Proton::Handler::MessagingHandler
+  def initialize(options = {})
+    require 'qpid_proton'
+
+    super()
+    @options = options
+
+    @url = @options.delete(:url)
+    @topics = @options.delete(:topics)
+    @test_connection = @options.delete(:test_connection)
+    @message_handler_block = @options.delete(:message_handler_block)
+  end
+
+  def on_start(event)
+    @conn = event.container.connect(@url, @options)
+    unless @test_connection
+      @topics.each { |topic| event.container.create_receiver(@conn, :source => "topic://#{topic}") }
+    end
+  end
+
+  def on_connection_opened(event)
+    # In case connection test was requested, close the connection immediately.
+    event.container.stop if @test_connection
+  end
+
+  def on_connection_error(_event)
+    raise MiqException::MiqInvalidCredentialsError, "Connection failed due to bad username or password"
+  end
+
+  def on_transport_error(_event)
+    raise MiqException::MiqHostError, "Transport error"
+  end
+
+  def on_message(event)
+    @message_handler_block.call(JSON.parse(event.message.body)) if @message_handler_block
+  end
+
+  def stop
+    @conn.close
+  end
+end
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
@@ -1,0 +1,35 @@
+class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Runner < ManageIQ::Providers::BaseManager::EventCatcher::Runner
+  def event_monitor_handle
+    unless @event_monitor_handle
+      options = @ems.event_monitor_options
+      options[:topics] = worker_settings[:topics]
+      @event_monitor_handle = ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream.new(options)
+    end
+    @event_monitor_handle
+  end
+
+  def reset_event_monitor_handle
+    @event_monitor_handle = nil
+  end
+
+  # Start monitoring for events. This method blocks forever until stop_event_monitor is called.
+  def monitor_events
+    event_monitor_handle.start do |event|
+      event_monitor_running
+      @queue.enq(event)
+    end
+  ensure
+    stop_event_monitor
+  end
+
+  def stop_event_monitor
+    @event_monitor_handle.try!(:stop)
+  ensure
+    reset_event_monitor_handle
+  end
+
+  def queue_event(event)
+    event_hash = ManageIQ::Providers::Nuage::NetworkManager::EventParser.event_to_hash(event, @cfg[:ems_id])
+    EmsEvent.add_queue('add', @cfg[:ems_id], event_hash)
+  end
+end
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
@@ -1,0 +1,45 @@
+class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream
+  include Vmdb::Logging
+
+  def self.test_amqp_connection(options = {})
+    # Ensure we just test the connection. AMQP channel will be established and
+    # started, however it will be immediately stopped.
+    options[:test_connection] = true
+    begin
+      stream = new(options)
+      stream.connection.run
+      true
+    rescue => e
+      _log.info("#{log_prefix} Failed connecting to ActiveMQ: #{e.message}")
+      raise
+    end
+  end
+
+  def self.log_prefix
+    "MIQ(#{self.class.name})"
+  end
+
+  def initialize(options = {})
+    require 'qpid_proton'
+
+    @options = options
+  end
+
+  def start(&message_handler_block)
+    _log.debug("#{self.class.log_prefix} Opening amqp connection using options #{@options}")
+    @options[:message_handler_block] = message_handler_block if message_handler_block
+    connection.run
+  end
+
+  def stop
+    @handler.stop
+  end
+
+  def connection
+    unless @connection
+      @handler = ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler.new(@options)
+      @connection = Qpid::Proton::Reactor::Container.new(@handler)
+    end
+    @connection
+  end
+end
--- a/app/models/manageiq/providers/nuage/network_manager/event_parser.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_parser.rb
@@ -1,0 +1,14 @@
+module ManageIQ::Providers::Nuage::NetworkManager::EventParser
+  def self.event_to_hash(event, ems_id)
+    event_type = "#{event['entityType']}_#{event['type'].downcase}"
+    {
+      :source     => "NUAGE",
+      :event_type => event_type,
+      :timestamp  => DateTime.strptime((event["eventReceivedTime"] / 1000).to_s, '%s').to_s,
+      :message    => event_type,
+      :vm_ems_ref => nil,
+      :full_data  => event.to_hash,
+      :ems_id     => ems_id
+    }
+  end
+end
--- a/app/models/manageiq/providers/nuage/network_manager.rb
+++ b/app/models/manageiq/providers/nuage/network_manager.rb
@@ -1,6 +1,7 @@
 class ManageIQ::Providers::Nuage::NetworkManager < ManageIQ::Providers::NetworkManager
   include SupportsFeatureMixin
-
+  require_nested :EventCatcher
+  require_nested :EventParser
   require_nested :RefreshParser
   require_nested :RefreshWorker
   require_nested :Refresher
@@ -20,5 +21,23 @@ class ManageIQ::Providers::Nuage::NetworkManager < ManageIQ::Providers::NetworkM
 
   def self.description
     @description ||= "Nuage Network Manager".freeze
+  end
+
+  def authentications_to_validate
+    at = [:default]
+    at << :amqp if has_authentication_type?(:amqp)
+    at
+  end
+
+  def supported_auth_types
+    %w(default amqp)
+  end
+
+  def supports_authentication?(authtype)
+    supported_auth_types.include?(authtype.to_s)
+  end
+
+  def self.event_monitor_class
+    ManageIQ::Providers::Nuage::NetworkManager::EventCatcher
   end
 end
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher.rb
@@ -1,0 +1,11 @@
+class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher < ::MiqEventCatcher
+  require_nested :Runner
+
+  def self.ems_class
+    ManageIQ::Providers::Nuage::NetworkManager
+  end
+
+  def self.settings_name
+    :event_catcher_nuage_network
+  end
+end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -14,3 +14,9 @@
   :nuage_network:
     :inventory_object_refresh: true
     :allow_targeted_refresh: true
+:workers:
+  :worker_base:
+    :event_catcher:
+      :event_catcher_nuage_network:
+        :topics:
+          - topic/CNAMessages
--- a/spec/factories/ext_management_system.rb
+++ b/spec/factories/ext_management_system.rb
@@ -21,4 +21,12 @@ FactoryGirl.define do
       ems.authentications << FactoryGirl.create(:authentication, cred)
     end
   end
+
+  factory :ems_nuage_network_with_authentication,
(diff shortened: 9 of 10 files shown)


gberginc:
@miq-bot add_label wip

The reason for WIP is that it requires Qpid Proton gem that is currently not properly available (it also requires host libs). Hopefully, the required version will be ready in the second half of September!

In order to test this, I am using the following gem in my main Gemfile
```
gem "qpid_proton", "~>0.17.0.1", :git => "https://github.com/gberginc/qpid_proton_gem"
```

Dry code review would be much appreciated to simplify merging afterwards.

@miha-plesko Would you be willing to make the first pass?

app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
```diff
@@ -0,0 +1,40 @@
...
+    reset_event_monitor_handle
+  end
+
+  def queue_event(event)
+    event_hash = ManageIQ::Providers::Nuage::NetworkManager::EventParser.event_to_hash(event, @cfg[:ems_id])
+    EmsEvent.add_queue('add', @cfg[:ems_id], event_hash)
+  end
+
+  def topic_names
+    "topic://topic/CNAMessages"
```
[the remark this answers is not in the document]
gberginc: Yes, I am discussing this with Nuage currently. I think using settings will be ideal, in particular because there are two at least two topics (events and alerts) that are of interest.

app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
```diff
@@ -0,0 +1,41 @@
...
+    event.container.create_receiver(conn, :source => @topic) unless @test_connection
+  end
+
+  def on_connection_opened(event)
+    # In case connection test was requested, close the connection immediately.
+    event.container.stop if @test_connection
+  end
+
+  def on_connection_error(event)
+    raise MiqException::MiqInvalidCredentialsError.new "Connection failed due to bad username or password"
```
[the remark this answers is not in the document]
gberginc: From what I have seen, the `event` does not contain additional information. I've seen CC comments and will address them

app/models/manageiq/providers/nuage/manager_mixin.rb
```diff
@@ -41,12 +41,46 @@ def translate_exception(err)
...
+    case auth_type.to_s
+    when 'default' then verify_api_credentials(options)
+    when 'amqp'    then verify_amqp_credentials(options)
+    end
+  end
+
+  def auth_url(protocol, server, port, version)
+    scheme = protocol == "ssl-with-validation" ? "https" : "http"
+    "#{scheme}://#{server}:#{port}/nuage/api/#{version}"
+  end
```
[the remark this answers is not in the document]
gberginc: No, I incorrectly moved it while introducing new functionalities

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,46 @@
+require 'qpid_proton'
+
+class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream
+  def self.test_amqp_connection(options = {})
+    # Ensure we just test the connection. AMQP channel will be established and
+    # started, however it will be immediately stopped.
+    options[:test_connection] = true
```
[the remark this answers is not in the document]
gberginc: Yes, unfortunately it works like this. In case the URL is incorrect, this will be noticed immediately when opening the connection, but in case just the `user:pass` is incorrect, we need to wait for the connection to be opened - otherwise the corresponding error event will be thrown.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  def self.log_prefix
+    "MIQ(#{self.class.name})"
+  end
+
+  def initialize(options = {})
+    @options           = options
+    @collecting_events = false
```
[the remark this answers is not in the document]
gberginc: Nah, that is your fault as I followed your [example](https://github.com/ManageIQ/manageiq-providers-vmware/blob/ba1583e269620784bdbe5e367e80fed61fc95f5d/app/models/manageiq/providers/vmware/cloud_manager/event_catcher/stream.rb#L36) 😜

gberginc:
Thanks @miha-plesko! All comments addressed.

CC now happy, but now rubocop says I have set everything on fire 😄.

@juliancheal the two lines identified as `unexpected token error` were changed from

```
@message_handler_block.call(JSON.parse(event.message.body)) if @message_handler_block
@event_monitor_handle.try!(:stop)
```

because codeclimate suggested the use of `safe navigation (&.)`. I am fine either way, depends mainly on what the target Ruby version is currently (safe navigation has been around since 2.3, but seems like rubocop is not configured for that).

gberginc:
@miq-bot add_label enhancement,events

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def start(&message_handler_block)
+    _log.debug("#{self.class.log_prefix} Opening amqp connection using options #{@options}")
+    @options[:message_handler_block] = message_handler_block if message_handler_block
+    connection.run
+  end
+
+  def stop
+    @handler.stop
```
miha-plesko: I understand `@handler.stop` sets `@stop_listening = true` which is then read in `on_message` function of the MessagingHandler. I'm wondering if this is sufficient - if no new message arrives after stopping, then the @handler will keep running.

I'm not familiar with qpid proton gem, but would it work if we use something like:

```ruby
@handler.interrupt
```

to stop the event catcher? See [this](https://github.com/gberginc/qpid_proton_gem/blob/6188f0732382fc7983f5c0aa8f6da379050d71fd/lib/messenger/messenger.rb#L399-L413) documentation. Please note, it's just an idea of mine, not sure if it's doable.

gberginc: Problem with this is that it's again part of the `messenger`, not the `reactor`. The reactor was chosen for this use case as it seems more lightweight as we are only listening to the events.

I understand your concern and will investigate the reactor to see if there is similar way to stop  the handler

gberginc: @miha-plesko Check [this commit](https://github.com/ManageIQ/manageiq-providers-nuage/pull/9/commits/f5707bc13f6a3142390abbe0236c3d6f1b097cc4) please. This now automatically closes the connection when stop is requested. Thanks for pointing out the problem.

gberginc:
@miha-plesko I've added a [patch](https://github.com/ManageIQ/manageiq-providers-nuage/pull/9/commits/9f1124be95b6ba406515285d1fd8fc9806b7c9a6) that uses event catcher settings to specify the topics of interest. This allows us to enable standard events as well as alarms from the config file, with no change in the code. With the previous commit that handles connection closing immediately, it makes it much more practical.

gberginc:
@juliancheal Do you perhaps have some time to review the code, despite the fact that we don't have the correct gem in place? Please note that the worker is not enabled by default (it's not yet added [here](https://github.com/ManageIQ/manageiq/blob/master/lib/workers/miq_worker_types.rb#L32)) so this PR does not actually break anything.

app/models/manageiq/providers/nuage/manager_mixin.rb
```diff
@@ -41,12 +41,41 @@ def translate_exception(err)
...
+      end
+
+      if (authentication = amqp.try(:authentication))
+        url = "#{authentication.userid}:#{authentication.password}@#{url}"
+      end
+
+      {
+        :ems                       => self,
+        :url                       => url,
+        :sasl_allow_insecure_mechs => true, # Only plain (insecure) mechanism currently supported
```
miha-plesko: Now that we've introduced settings.yaml, we could probably also move `:sasl_allow_insecure_mechs` decision in there.

gberginc: Insecure will come from the AMQP settings from the UI (you probably remember the "new provider" where you have an option for secure/not secure).

The reason to have this hardcoded is simply because Nuage supposedly only uses amqp and not amqps - thus, for now even the UI will not allow ssl.

miha-plesko: Ah, yes, you are right. Thanks for the explanation!

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
miha-plesko: I like it better now, thanks for this change.

sergio-ocon:
How is the status of the library needed?
Would this work in other places?

gberginc:
I haven't seen the update from qpid_proton - will post a question.

From what I was able to test, this PR is now in a shape that it should allow merging without any disruption in the existing provider. Eventing will not be enabled unless a change in core is also made.

juliancheal:
> Thanks @miha-plesko! All comments addressed.
> 
> CC now happy, but now rubocop says I have set everything on fire 😄.
> 
> @juliancheal the two lines identified as unexpected token error were changed from
> 
> @message_handler_block.call(JSON.parse(event.message.body)) if @message_handler_block
> @event_monitor_handle.try!(:stop)
> because codeclimate suggested the use of safe navigation (&.). I am fine either way, depends mainly on what the target Ruby version is currently (safe navigation has been around since 2.3, but seems like rubocop is not configured for that).

I think we are still supporting 2.2.2, so wouldn't be able to use `&.`. I remember this coming up with a PR from Lenovo. @Fryguy would would you say.

juliancheal:
I'm good with this PR, except the `&.`, but want to see if @Fryguy agrees.

What's stopping the correct gem being in place to be used by this PR?

gberginc:
@juliancheal [this](https://github.com/apache/qpid-proton/pull/116#issuecomment-331201529) is the latest response from Qpid team. Problem is that current version ([0.17](https://rubygems.org/gems/qpid_proton/versions/0.17.0)) has a bug preventing Nuage provider to connect to ActiveMQ. We have fixed this (the above PR explains in some more detail) and are currently using a custom Gem with the patch applied.

As written above, the event catcher currently does not affect anything because it is not enabled by default. I am thus fine removing the WIP in case you are also fine accepting the code in such state. Once new Gem is released, we will be able to enable it in the list of workers (core repo).

juliancheal:
@gberginc thanks for the clarification 👍

sergio-ocon:
What happened to the underlying libraries needed for this to work?
Do we need changes in the build to make this work?

gberginc:
@juliancheal Since I have reverted the use of "safe navigation", this again causes codeclimate to fail, however I believe is is still much better than those explosive rubocop warnings about older Ruby versions. Also, it's in line with the code of other similar providers.

I am now also removing the wip label as it should be safe for the merge.

@miq-bot remove_label wip

app/models/manageiq/providers/nuage/network_manager/event_parser.rb
```diff
@@ -0,0 +1,13 @@
+module ManageIQ::Providers::Nuage::NetworkManager::EventParser
+  def self.event_to_hash(event, ems_id)
+    {
+      :source     => "Nuage",
```
blomquisg: For consistency sake, please change to "NUAGE".

gberginc:
Thank you very much @blomquisg for taking the time to review the code. I've changed the source as you suggested and am now waiting for @durandom.

sergio-ocon:
Code climate is complaining for not using &., but we asked them to do so before to be able to use ruby 2.2

@durandom ?

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,45 @@
+require 'qpid_proton'
```
durandom: the `qpid_proton` gem is nowhere included. AFAIK we dont use it yet in the core.

@kbrock when / will we include it in the cores Gemfile?

If not you'll have to add it to the providers gemspec

durandom: @gberginc  if you add a spec, that will trigger this code path, I guess it will throw an exception

app/models/manageiq/providers/nuage/network_manager/event_parser.rb
```diff
@@ -0,0 +1,13 @@
+module ManageIQ::Providers::Nuage::NetworkManager::EventParser
+  def self.event_to_hash(event, ems_id)
+    {
+      :source     => "NUAGE",
+      :event_type => "#{event['entityType']}_#{event['type'].downcase}",
+      :timestamp  => DateTime.strptime((event["eventReceivedTime"] / 1000).to_s, '%s').to_s,
+      :message    => event.to_hash,
```
durandom: I think `message` should rather be a single string and `full_data` contains the complete hash.

app/models/manageiq/providers/nuage/network_manager/event_parser.rb
```diff
@@ -0,0 +1,13 @@
...
+      :vm_ems_ref => nil,
+      :full_data  => event,
```
durandom: I guess this will be a hash?
But maybe, just to be sure, when something in qpid changes and they refactor to use a object instead, you'll want to make this `event.to_hash`

gberginc:
@durandom Thanks!

You were not in the emails we are exchanging with Satoe about Qpid gem. Two things are required: first, few [system packages](https://github.com/ManageIQ/manageiq-appliance-build/pull/228/files), and then install the gem from a forked [Qpid repo](https://github.com/gberginc/qpid_proton_gem). Problem with the latter is that Qpid has not released an update that is mandatory for this to work with Nuage.

I have prepared a [simple demo repo](https://github.com/gberginc/nuage_amqp_client) to demonstrate how it works.

Can you please advise what I should do? Should I propose to introduce the gem into the main Gemfile?

gberginc:
@durandom The [latest patch](https://github.com/ManageIQ/manageiq-providers-nuage/pull/9/commits/5f07e22955878ca078c52eda5305ab5324ec4c78) addresses the comments in the code.

Now I need to resolve the Qpid thing...

If other parts are fine, is it possible to get this in and then submit specs (which we are already working on) separately?

durandom:
> Can you please advise what I should do? Should I propose to introduce the gem into the main Gemfile?

@kbrock @Fryguy how should we handle the qpid dependency?
I'd say include it in this providers gemspec

gberginc:
Problem with provider's gemspec is that bundler does not allow github repositories. Thus, currently, the only way it is to get it inside main Gemfile. Once Qpid gets new build in RubyGems, provider's gemspec is the target.

juliancheal:
> Code climate is complaining for not using &., but we asked them to do so before to be able to use ruby 2.2

We asked Code Climate you mean? Yeah it's just for now while we support 2.2

juliancheal:
What's the status of the Qpid dependency?

gberginc:
The appliance-build has green light for [this PR](https://github.com/ManageIQ/manageiq-appliance-build/pull/228/files) (confirmed via email so I expect it to be merged soon).

The gem is still officially at 0.17, however the authors are working on releasing 0.18.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
Fryguy: > If not you'll have to add it to the providers gemspec

It should be included in the provider's gemspec regardless, because the provider should always declare it's dependencies.  This way if we, say, have to bump the version in the main repo, we can ensure we pick a version that satisfies the dependencies of this plugin

Fryguy:
Merged https://github.com/ManageIQ/manageiq-appliance-build/pull/228

Fryguy:
@gberginc What are you changing in the gem that's requiring you to use a git repo?

gberginc:
@Fryguy I can add the `qpid_proton` gem into the Nuage gemspec and it will pass most of the tests we have already in place offline. The problem with the current published version (0.17) is that it does not work with the authentication method Nuage is using (i.e. PLAIN). This means that using the current version, it will not connect to Nuage AMQP.

Therefore, I created a separated repo to host the 0.17 + the patch that was implemented based on the report I have provided to Qpid team.

Fryguy:
Ah ok...I misunderstood...from the way the appliance-build PR went, I thought that PLAIN was just a matter of adding the cyrus_sasl_plain package.

gberginc:
I've added a second PR (#10) that includes `qpid_proton` in gemspec (0.17, but for the tests this does not matter). However, it does not get installed due to a [missing system library](https://travis-ci.org/ManageIQ/manageiq-providers-nuage/jobs/283036285#L832).

Will the change to `manageiq-appliance-build` resolve this at some point?

durandom:
> Will the change to manageiq-appliance-build resolve this at some point?

afaik, the appliance stuff is not included for travis builds.

So, from my understanding, adding qpid and the .travis.yml changes to this PR here would make it mergeable.

sergio-ocon:
@gberginc Have those changes been done?

sergio-ocon:
The gem was supposed to be released already.
I've replied . to your request and Philipp's about this.

gberginc:
Thank you @sergio-ocon! Please, also see my last question in #11. I really would appreciate feedback to resolve this problem.

chargio:
@durandom @Fryguy Can you please have another look at this? With the PR that have been merged for the qpid gem, we should be ready to merge this.
If there is anything outstanding, please let us know so we can fix it this week.

Thanks

gberginc:
@durandom Thank you very much for you review!

The way I understand gemspec, it's not possible to add a gem from GitHub repository into gemspec. Once the gem will be released officially, I will add that as a dependency in Nuage provider.

I will also add the guides, yes. I just need to make sure to try every platform (Fedora, Ubuntu and mac OS).

gberginc:
@miq-bot assign @Fryguy 

This is now green based on the latest changes.

chargio:
@gberginc Is there any pending comment to address or are all them addressed?

juliancheal:
Seems like everything is done, as quid gem is installed now via other prs. @Fryguy are all your issues resolved? I'll merge this if so.

gberginc:
@chargio as far as I am concerned this PR is ready. Others, simpler, will follow.

app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
```diff
@@ -0,0 +1,41 @@
+require 'qpid_proton'
```
Fryguy: This cannot be in the top-level context.  As the gem is not required to be installed, this class will blow up on load (even if the worker is not started).  You can avoid this by just putting the require in the initialize call.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
Fryguy: Same...this still cannot be at the top-level context

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,45 @@
...
+
+class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream
+  include Vmdb::Logging
+
+  def self.test_amqp_connection(options = {})
+    # Ensure we just test the connection. AMQP channel will be established and
+    # started, however it will be immediately stopped.
+    options[:test_connection] = true
+    begin
+      stream = ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream.new(options)
```
Fryguy: Minor, but you don't need to retype out the class name here...

```ruby
stream = new
stream.connection.run
```

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
Fryguy: I don't think you even need this as long as the MessagingHandler does the require in it's initialize method.

gberginc:
@Fryguy just responding here as well in case you miss the gitter comment. Thanks for the review.

chargio:
Thanks @gberginc. @Fryguy it is supposed to be approved by every body now

Fryguy:
Need this one addressed: https://github.com/ManageIQ/manageiq-providers-nuage/pull/9#discussion_r146873094

After that, this is good to go!

juliancheal:
> Need this one addressed: #9 (comment)
> 
> After that, this is good to go!

Excellent, as soon as that change is in I'll press the big green button (which is actually grey) /cc @gberginc

juliancheal:
Cool beans. Just waiting on Travis.

chargio:
🥇

PR #10, opened 2017-10-04
Add Nuage manager and event catcher specs

gberginc:
With this patch we are adding specs to test the Nuage network manager,
event catcher and event parser.

### DIFF
--- a/spec/models/manageiq/providers/nuage/network_manager/event_catcher/alarm_delete.json
+++ b/spec/models/manageiq/providers/nuage/network_manager/event_catcher/alarm_delete.json
@@ -1,0 +1,38 @@
+{
+  "assoicatedEvent": true,
+  "diffMap": null,
+  "enterpriseName": null,
+  "entities": [
+    {
+      "ID": "5911286f-030c-43c2-8fef-4b01fdef40df",
+      "acknowledged": false,
+      "alarmedObjectID": "a6ef5485-c8fb-4509-9e70-51e4262a2abf",
+      "children": null,
+      "creationDate": 1506659170000,
+      "description": "Gateway [201.26.92.41:NSGBR] was disconnected from the controller.",
+      "enterpriseID": "76046673-d0ea-4a67-b6af-2829952f0812",
+      "entityScope": "GLOBAL",
+      "errorCondition": 4707,
+      "externalID": null,
+      "lastUpdatedBy": "43f8868f-4bc1-472c-9d19-533dcfcb1ee0",
+      "lastUpdatedDate": 1506659170000,
+      "name": "NSGBR 201.26.92.41 Gateway with system-id [201.26.92.41] is disconnected",
+      "numberOfOccurances": 1,
+      "owner": "43f8868f-4bc1-472c-9d19-533dcfcb1ee0",
+      "parentID": "a6ef5485-c8fb-4509-9e70-51e4262a2abf",
+      "parentType": "nsgateway",
+      "reason": "Gateway with system-id [201.26.92.41] is disconnected",
+      "severity": "MAJOR",
+      "targetObject": "nsgateway",
+      "timestamp": 1506659170849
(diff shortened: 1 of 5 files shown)


gberginc:
@miq-bot assign @juliancheal 

This PR is now much simpler as we have the other ones merged. There is one issue that I would like to explicitly point out.

[This require](https://github.com/ManageIQ/manageiq-providers-nuage/pull/10/files#diff-93830fa29d616f7c87903d08b5b1b29aR6). As you remember from yesterday we moved these `require`s into methods to prevent breaking the code. This worked for the parser itself, however it causes problems in the specs, because of [this](https://github.com/ManageIQ/manageiq-providers-nuage/blob/master/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb#L1) -- the implementation extends a class from Qpid. Since spec is ran in isolation, I need to introduce the mandatory require. Let me know if `spec_helper` is the correct place.

@miq-bot add_label test, events

spec/models/manageiq/providers/nuage/network_manager/event_catcher/alarm_delete.json
```diff
@@ -0,0 +1,38 @@
+{
```
Fryguy: Minor, but you might want to move this into a directory named `spec/models/manageiq/providers/nuage/network_manager/event_catcher/data` (we use data as the name in other places).  This way if you add more data files, they are contained.

spec/models/manageiq/providers/nuage/network_manager/event_parser_spec.rb
```diff
@@ -0,0 +1,15 @@
+describe ManageIQ::Providers::Nuage::NetworkManager::EventParser do
+  context ".event_to_hash" do
+    it "parses vm_ems_ref into event" do
+      ems = FactoryGirl.create(:ems_nuage_network)
+      message = JSON.parse(File.read(File.join(File.dirname(__FILE__), "/event_catcher/alarm_delete.json")))
```
Fryguy: Prefer `__dir__` over `File.dirname(__FILE__)`

spec/models/manageiq/providers/nuage/network_manager/event_catcher/alarm_delete.json
Fryguy: Oh nvm...I thought the specs files and data would be mixed together (which still *might* happen if you create a new thing namespaced under event_catcher), but this *is* the data directory, so this is fine.

spec/models/manageiq/providers/nuage/network_manager_spec.rb
```diff
@@ -21,5 +38,98 @@
         @ems.verify_credentials(:unsupported)
       end.to raise_error(MiqException::MiqInvalidCredentialsError)
     end
+
+    context 'AMQP connection' do
+      before :each do
```
Fryguy: :each is redundant, so you can remove it.

spec/models/manageiq/providers/nuage/network_manager_spec.rb
```diff
@@ -21,5 +38,98 @@
...
+
+      it 'handles connection errors' do
+        allow(@conn).to receive(:run).and_raise(StandardError, 'connection error')
+        expect { @ems.verify_credentials(:amqp) }.to raise_error(StandardError)
+      end
+    end
+  end
+
+  context 'translate_exception' do
+    before :each do
```
Fryguy: Same.

gberginc:
@Fryguy comments addressed in the spec. Regarding `:each` I have now removed all of them, even in parts that were previously not touched by this PR. I want to avoid any future copy-paste stupidities.

gberginc:
@juliancheal All of the comments have been addressed; @blomquisg approved it so I think it's safe to merge it.

PR #11, opened 2017-10-04
Install qpid_proton for Travis

gberginc:
The worker used by Travis does not have the most current version of the Qpid Proton installed, so we need to manually install it from source. This patch also adds a simple spec that chects that Qpid gem is properly installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 rvm:
 - '2.3.3'
 - '2.4.1'
-sudo: false
+sudo: required
 cache: bundler
 env:
   global:
@@ -11,6 +11,7 @@ env:
   - RUBY_GC_HEAP_GROWTH_FACTOR=1.25
 addons:
   postgresql: '9.4'
+before_install: sudo bin/qpid_install.sh
 install: bin/setup
 after_script: bundle exec codeclimate-test-reporter
 notifications:
--- a/bin/qpid_install.sh
+++ b/bin/qpid_install.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+# Ensure correct Ruby environment is used.
+. ~/.rvm/scripts/rvm
+rvm use $TRAVIS_RUBY_VERSION
+
+# We have to extract the major and minor ruby versions so that we can set proper GEM_HOME
+ruby_version=`echo $TRAVIS_RUBY_VERSION | egrep -o '[[:digit:]]\.[[:digit:]]'`
+
+# As we are installing the Gem built from source, we have to set the proper
+# GEM_HOME. This will ensure that the gem will be installed in a place where it
+# will be picked by bundler.
+export GEM_HOME=/home/travis/build/ManageIQ/manageiq-providers-nuage/vendor/bundle/ruby/$ruby_version.0
+
+# Install the dev dependencies for building Qpid proton system library.
+sudo apt-get install -y gcc cmake cmake-curses-gui uuid-dev
+sudo apt-get install -y libssl-dev
+sudo apt-get install -y libsasl2-2 libsasl2-dev
+sudo apt-get install -y swig
+
+# Get the latest Qpid Proton source
+cd $HOME/build
+git clone https://github.com/apache/qpid-proton.git
+cd qpid-proton
+
+# There is a strange dependency on JSON that we need to change to version of
+# at least 2; otherwise there will be a conflict.
+sed -i .bak -e 's/ 0/ 2/' proton-c/bindings/ruby/qpid_proton.gemspec.in
+
+# Configure the source of Qpid Proton.
+mkdir build
+cd build
+cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DSYSINSTALL_BINDINGS=OFF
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -11,8 +11,14 @@ env:
   - RUBY_GC_HEAP_GROWTH_FACTOR=1.25
 addons:
   postgresql: '9.4'
-install: bin/setup
+install: 
+  - sudo bin/qpid_install.sh
+  - gem install specific_install
+  - gem specific_install https://github.com/xlab-si/qpid_proton_gem
```
Fryguy: Is this effectively a straight extraction of  https://github.com/apache/qpid-proton/tree/master/proton-c/bindings/ruby ?

gberginc:
@Fryguy @juliancheal I have finally managed to alter the way Travis installed all the necessary dependencies in order to pass Qpid Proton tests. I am now building the code directly from source code  (both system libraries and Ruby Gem) as this will be the closest to what we will eventually come once the Gem is released.

@Fryguy You asked me about the `specific_install` that was present in one of my attempts. Yes, that was what I was trying to use, however it worked only for one of the Ruby versions (I think it was 2.4.1, but not for 2.3.3). Therefore, I implemented that without the separate fork.

I really hope this will serve for the time being as I already have other specs ready but do not want to overcomplicate this simple PR with too many things.

juliancheal:
LGTM 👍, but I'll wait for @Fryguy's thoughts before merge.

gberginc:
Thanks @juliancheal!

Fryguy:
@gberginc Glad to see you got this green!  I'll take a look in a bit.

.travis.yml
```diff
@@ -11,7 +11,9 @@ env:
   - RUBY_GC_HEAP_GROWTH_FACTOR=1.25
 addons:
   postgresql: '9.4'
-install: bin/setup
+install: 
+  - sudo bin/qpid_install.sh
```
Fryguy: Prefer putting this in a `before_install:` section (that section is where dependencies are usually installed, which this is).

Fryguy:
@gberginc What is the timeline for an 0.18 release of qpid_proton?  I'm just trying to get a gauge on how long we would need to carry a custom build.

gberginc:
@Fryguy I've moved the installation of Qpid proton into `before_install` - still green :smile:.

Regarding the timeline, it is really hard to say. It was initially mid September, then moved to beginning of October. I believe it should be soon as they are already releasing beta versions of system libs, but can't say for sure. Even when 0.18 is released, system libs will have to be built into the worker somehow - do you perhaps cache built workers?

Fryguy:
> system libs will have to be built into the worker somehow

In the appliances, this happens in the nightly builds.  For developers, we would need to let them know how to upgrade / install if they are doing local development.  Unfortunately, because the qpid_proton gem needs compilation, there's no way to "delay build" it

gberginc:
I intend to update the documentation after we are through with the main chunk of code. For CentOS EPEL will be sufficient, for Ubuntu/Debian, current build script is a starting point (unfortunately, there are no latest packages for e.g. Ubuntu 16.04, current LTS).

Fryguy:
This PR looks good to me code-wise, but I'm very hesitant on merging it without proper documentation for all developers.  The moment it gets merged, every developer will automatically bring in the qpid_proton dependency, and it *must* build on their machine (which is like 95% Fedora and Mac).  Additionally, considerations must be made for the nightly appliance builds.  Since this PR locks the version to 0.18, and since that isn't in rubygems, then they *all* will need to some sort of manual approach.

To move forward, and to get Travis in a working position, please remove the changes to the gemspec in this PR.  While you will have to carry those changes in the next PR, this will at least let us merge this one and get Travis working.

gberginc:
@Fryguy I have now removed the `qpid_proton` from gemspec, however the tests are now failing (as expected).

Was this your intention?

Fryguy:
You'll have to remove the require from the spec.  Technically that spec doesn't need it anyway.

gberginc:
I see. I thought that the plan was to enable Travis tests for Qpid-related changed. The reason for me having to build Qpid from source was that Travis workers are based on Ubuntu which has no official release of latest Qpid version.

On the other hand, for Fedora, EPEL provides packages for 0.17, which is sufficient for most of the cases, i.e. developers will only have to install qpid_proton 0.17 system lib and ruby gem. The same goes for Mac users where there is Homebrew package (requires some work, but quite trivial).

This is now green, however I am honestly out of ideas how to proceed with the merges, in particular with the code freeze getting so close 😞. Any suggestion would be much appreciated.

gberginc:
@Fryguy Would you consider adding the gem into core Gemfile a better alternative? This would mean that I would reference the Gem from my GitHub repository, until it is released officially. This Gem would include the changes necessary for the Nuage provider.

Thus, travis would still build Qpid system libs manually, but the gem would be downloaded automatically.

Fryguy:
> Would you consider adding the gem into core Gemfile a better alternative? 

Yeah, I think that's a better approach, especially since you are going to have a git-based gem anyway.

PR #12, opened 2017-10-06
Provide extensive unit tests for legacy refresher

miha-plesko:
This commit is a preparation to be able to support graph inventory refresh with the next commit. Basically we provide unit tests for refresh parser so that we'll be able to assure that the new graph inventory refresh works identically to the current (legacy) one.

Writing tests we've also discovered that there is a bug in current parser - the network groups were not associated correctly with the security groups. Fixed. God bless the unit tests! :D

@miq-bot add_label enhancement
@miq-bot assign @juliancheal 

/cc @gberginc

PR #13, opened 2017-10-06
Implement graph inventory refresh for network manager

miha-plesko:
With this commit we implement whatever is needed for the fancy new graph inventory refresh to step in. For now we keep legacy refresher also in place (switchable in settings.yml), but the plan is to migrate to the new refresher completely.

Exactly same unit tests that were passing for legacy refresher are now passing for both the legacy refresher and the new graph refresher.

Depends on: https://github.com/ManageIQ/manageiq-providers-nuage/pull/12
Depends on: https://github.com/ManageIQ/manageiq/pull/16136

@miq-bot assign @juliancheal 

/cc @gberginc

### DIFF
--- a/app/models/manageiq/providers/nuage/inventory/persister.rb
+++ b/app/models/manageiq/providers/nuage/inventory/persister.rb
@@ -1,0 +1,26 @@
+class ManageIQ::Providers::Nuage::Inventory::Persister < ManagerRefresh::Inventory::Persister
+  require_nested :NetworkManager
+
+  protected
+
+  def network
+    ManageIQ::Providers::Nuage::InventoryCollectionDefault::NetworkManager
+  end
+
+  def targeted
+    false
+  end
+
+  def strategy
+    nil
+  end
+
+  def shared_options
+    settings_options = options[:inventory_collections].try(:to_hash) || {}
+
+    settings_options.merge(
+      :strategy => strategy,
+      :targeted => targeted,
+    )
+  end
+end
--- a/app/models/manageiq/providers/nuage/network_manager/refresher.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/refresher.rb
@@ -9,5 +9,38 @@ module ManageIQ::Providers
     def post_process_refresh_classes
       []
     end
+
+    def collect_inventory_for_targets(ems, targets)
+      log_header = format_ems_for_logging(ems)
+      targets_with_data = targets.collect do |target|
+        target_name = target.try(:name) || target.try(:event_type)
+
+        _log.info("#{log_header} Filtering inventory for #{target.class} [#{target_name}] id: [#{target.id}]...")
+
+        if refresher_options.try(:[], :inventory_object_refresh)
+          inventory = ManageIQ::Providers::Nuage::Builder.build_inventory(ems, target)
+        end
+
+        _log.info("#{log_header} Filtering inventory...Complete")
+        [target, inventory]
+      end
+
+      targets_with_data
+    end
+
+    def parse_targeted_inventory(ems, _target, inventory)
+      log_header = format_ems_for_logging(ems)
+      _log.debug("#{log_header} Parsing inventory...")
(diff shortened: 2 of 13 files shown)


miha-plesko:
I'm marking this PR as WIP since it depends on the other PR to be merged. Other than that, the code here is fully functional so I kindly ask @gberginc for a first pass of review.

app/models/manageiq/providers/nuage/network_manager/refresher.rb
```diff
@@ -9,5 +9,37 @@ def parse_legacy_inventory(ems)
     def post_process_refresh_classes
       []
     end
+
+    def collect_inventory_for_targets(ems, targets)
+      targets_with_data = targets.collect do |target|
+        target_name = target.try(:name) || target.try(:event_type)
+
+        _log.info("Filtering inventory for #{target.class} [#{target_name}] id: [#{target.id}]...")
```
gberginc: Despite the fact that other providers also doesn't seem to have `log_header` here, I suggest you add it.

spec/models/manageiq/providers/nuage/network_manager/vcr_specs/refresher_spec.rb
```diff
@@ -2,6 +2,19 @@
   ALL_REFRESH_SETTINGS = [
     {
       :inventory_object_refresh => false
+    },
+    {
+      :inventory_object_refresh => true
```
gberginc: I haven't seen an example of just using the `inventory_object_refresh => true` setting. Do you perhaps know to what it defaults (the next one?)

Ladas:
@miha-plesko ```:saver_strategy => :concurrent_safe, :saver_strategy => :concurrent_safe_batch``` will not work until we have unique indexes in our DB. That is planned for 5.0 release now and these will be used for Streaming refresh (finally not loading everything into memory :-)) 

You might want to test what we test in AWS https://github.com/Ladas/manageiq-providers-amazon/blob/a39ed3dfc63bbdab1687f7bbfa5f396f06442218/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb#L11

While this setting is being the fastest and default in AWS https://github.com/Ladas/manageiq-providers-amazon/blob/a39ed3dfc63bbdab1687f7bbfa5f396f06442218/spec/models/manageiq/providers/amazon/aws_refresher_spec_common.rb#L24

app/models/manageiq/providers/nuage/inventory/persister.rb
```diff
@@ -0,0 +1,42 @@
+class ManageIQ::Providers::Nuage::Inventory::Persister < ManagerRefresh::Inventory::Persister
+  require_nested :NetworkManager
+
+  # TODO(lsmola) figure out a way to pass collector info, probably via target, then remove the below
+  attr_reader :collector
```
Ladas: unless we need the collector inside of the initialize_inventory_collections method, we can drop this and the initialize method. (will make it easier for me do do the refactoring later :-))

Ladas: yeah, I don't see we need this. Actually I might have already done that TODO work, just need to delete the TODOs :-D

miha-plesko:
Thanks for the review @Ladas and explanation regarding `:saver_strategy`. I've ended up using the very same refresher configuration options like they are used for Amazon provider, and everything works like a charm for Nuage 🎆 🍾 🥂  .

I've just rebased and repushed this whole PR and made it undependent of the pending [PR](https://github.com/ManageIQ/manageiq/pull/16136) on miq core. Perhaps we could merge this without waiting the core by introducing the [TODO](https://github.com/ManageIQ/manageiq-providers-nuage/pull/13/files#diff-3c8dfc503df949f7b7b9e1b88c5a8642R54)? I'm all on fire to implement the targeted refresh, but can't do it until Graph Refresh is in place...

BTW, I like the new graph refresher approach very much since the code is awesomely readable now that collector, parser and persister are split 👍

PR #15, opened 2017-10-13
Let Refresher extend the new ManagerRefresher

miha-plesko:
Recently the ManagerRefresher was introduced in the core repository. By extending it, we cleanup our Nuage::NetworkManager::Refresher so that it no longer implements the two copy-pasted functions:

- collect_inventory_for_targets
- parse_targeted_inventory

@miq-bot assign @Ladas 
@miq-bot add_label refactoring


miha-plesko:
@Ladas as promised I'm sending in the cleanup now that the ManagerRefresher has been accepted in the core repo.

PR #16, opened 2017-10-17
Rspec test for vsd_client.rb

gasper-vrhovsek:
Contains some basic test cases for the vsd client. Tests
ok responses for common parameters, expects errors on
wrong username and password.

### DIFF
--- a/spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
+++ b/spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
@@ -1,0 +1,392 @@
+require 'date'
+describe ManageIQ::Providers::Nuage::NetworkManager::VsdClient do
+  include Vmdb::Logging
+
+  UUID_REGEXP = /([0-9a-z]{8})-(([0-9a-z]{4})-){3}([0-9a-z]{12})/
+  ROUTE_PROPERTY_REGEXP = /([0-9]{5}):([0-9a-z]{4,5})/
+  IP_ADDR_REGEXP = /^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$/
+  MAC_ADDR_REGEXP = /^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$/
+
+  ENTERPRISE_ID = '6abac3ad-a05d-4b93-9556-4ba5010fb13b'.freeze
+  DOMAIN_ID = 'd0c3274c-397a-4173-8981-bfd2f99ef8c6'.freeze
+  ZONE_ID = '76ac549a-5843-4f1b-ac90-b21ff6edc2a4'.freeze
+  SUBNET_ID = '4da803f7-c0d2-4beb-b02e-341ea77e377f'.freeze
+  POLICY_GROUP_ID = 'fadd09c4-9fea-46ec-8342-73f1b6a4df74'.freeze
+
+  before(:each) do
+    @userid = Rails.application.secrets.nuage_network.try(:[], 'userid') || 'NUAGE_USER_ID'
+    @password = Rails.application.secrets.nuage_network.try(:[], 'password') || 'NUAGE_PASSWORD'
+    @hostname = Rails.application.secrets.nuage_network.try(:[], 'host') || 'nuagenetworkhost'
+
+    # Ensure that VCR will obfuscate the basic auth
+    VCR.configure do |c|
+      # workaround for escaping host
+      c.before_playback do |interaction|
+        interaction.filter!(CGI.escape(@hostname), @hostname)
+        interaction.filter!(CGI.escape('NUAGE_NETWORK_HOST'), 'nuagenetworkhost')
+      end
+      c.filter_sensitive_data('NUAGE_NETWORK_AUTHORIZATION') { Base64.encode64("#{@userid}:#{@password}").chomp }
+    end
+  end
+
+  def vcr_play(cassette, method, args: [])
+    VCR.use_cassette("#{described_class.parent.name.underscore}/vsd_client/#{cassette}") do
+      @vsd_client.send(method, *args)
+    end
+  end
+
+  context "when login successful" do
+    before(:each) do
+      VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/login') do
+        @vsd_client = described_class.new("https://#{@hostname}:8443/nuage/api/v5_0", @userid, @password)
+      end
+    end
+
+    it "should return valid non empty response for get_enterprises" do
+      enterprises = vcr_play('enterprises', :get_enterprises)
+      assert_object_not_empty(enterprises)
+      assert_enterprises(enterprises)
+    end
+
+    it "should return valid non empty response for get_enterprise by id" do
+      enterprise = vcr_play('enterprise', :get_enterprise, :args => ENTERPRISE_ID)
+      assert_object_not_empty(enterprise)
+      assert_enterprise(enterprise)
+    end
+
+    it "should return valid non empty response for domains" do
+      domains = vcr_play('domains', :get_domains)
+      assert_object_not_empty(domains)
+      assert_domains(domains)
+    end
+
+    it "should return valid non empty response for domain by id" do
+      domain = vcr_play('domain', :get_domain, :args => DOMAIN_ID)
+      assert_object_not_empty(domain)
+      assert_domain(domain)
+    end
+
+    it "should return valid non empty response for domains for enterprise_id" do
+      domains = vcr_play('domains_for_enterprise_id', :get_domains_for_enterprise, :args => ENTERPRISE_ID)
+      assert_object_not_empty(domains)
+      assert_domains(domains)
+    end
+
+    it "should return valid non empty response for zones" do
+      zones = vcr_play('zones', :get_zones)
+      assert_object_not_empty(zones)
+      assert_zones(zones)
+    end
+
+    it "should return valid non empty response for zone by id" do
+      zone = vcr_play('zone', :get_zone, :args => ZONE_ID)
+      assert_object_not_empty(zone)
+      assert_zone(zone)
+    end
+
+    it "should return valid non empty response for subnets" do
+      subnets = vcr_play('subnets', :get_subnets)
+      assert_object_not_empty(subnets)
+      assert_subnets(subnets)
+    end
+
+    it "should return valid non empty response for subnet by id" do
+      subnet = vcr_play('subnet', :get_subnet, :args=>SUBNET_ID)
+      assert_object_not_empty(subnet)
+      assert_subnet(subnet)
+    end
+
+    it "should return valid non empty response for subnets for domain" do
+      subnets = vcr_play('subnets_for_domain', :get_subnets_for_domain, :args => DOMAIN_ID)
+      assert_object_not_empty(subnets)
+      assert_subnets(subnets)
+    end
+
+    it "should return valid non empty response for vms" do
+      vms = vcr_play('vms', :get_vms)
+      assert_object_not_empty(vms)
+      assert_vms(vms)
+    end
+
+    it "should return valid non empty response for policy_groups" do
+      policy_groups = vcr_play('policy_groups', :get_policy_groups)
+      assert_object_not_empty(policy_groups)
+      assert_policy_groups(policy_groups)
+    end
+
+    it "should return valid non empty response for policy_group by id" do
+      policy_group = vcr_play('policy_group', :get_policy_group, :args => POLICY_GROUP_ID)
+      assert_object_not_empty(policy_group)
+      assert_policy_group(policy_group)
+    end
+
+    it "should return valid non empty response for policy_groups for domain" do
+      policy_groups = vcr_play('policy_groups_for_domain', :get_policy_groups_for_domain, :args=>DOMAIN_ID)
+      assert_object_not_empty(policy_groups)
+      assert_policy_groups(policy_groups)
+    end
+  end
+
+  context "when login fails" do
+    it "should fail on wrong password" do
+      VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/wrong_pass') do
+        expect do
+          described_class.new("https://#{@hostname}:8443/nuage/api/v5_0", @userid, 'wrong_password')
+        end.to raise_error(MiqException::MiqInvalidCredentialsError)
+      end
+    end
+
+    it "should fail on wrong username" do
+      VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/wrong_user') do
+        expect do
+          described_class.new("https://#{@hostname}:8443/nuage/api/v5_0", 'wrong_user', @password)
+        end.to raise_error(MiqException::MiqInvalidCredentialsError)
+      end
+    end
+
+    it "should fail on wrong hostname" do
+      VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/wrong_hostname') do
+        expect do
+          described_class.new("https://wronghost:8443/nuage/api/v5_0", @userid, @password)
+        end.to raise_error(SocketError)
+      end
+    end
+  end
+
+  def assert_object_not_empty(object)
+    expect(object.length).to be > 0
+    expect(object.length).to be_truthy
+  end
+
+  def assert_enterprises(enterprises)
+    expect(enterprises.count).to be > 0
+    assert_enterprise(enterprises.first)
+  end
+
+  def assert_enterprise(enterprise)
+    assert_time_fields(enterprise, %w(lastUpdatedDate creationDate))
+    assert_identifier_fields(enterprise, %w(ID lastUpdatedBy owner enterpriseProfileID
+                                            enterpriseProfileID receiveMultiCastListID
+                                            associatedGroupKeyEncryptionProfileID
+                                            associatedEnterpriseSecurityID associatedKeyServerMonitorID))
+    assert_string_fields(enterprise, %w(name))
+    assert_integer_fields(enterprise, %w(floatingIPsQuota floatingIPsUsed dictionaryVersion customerID))
+    assert_boolean_property_fields(enterprise, %w(allowTrustedForwardingClass allowAdvancedQOSConfiguration
+                                                  allowGatewayManagement enableApplicationPerformanceManagement
+                                                  LDAPEnabled LDAPAuthorizationEnabled BGPEnabled))
+    assert_entity_scope(enterprise['entityScope'])
+    assert_enabled_disabled_inherited_fields(enterprise, %w(encryptionManagementMode))
+  end
+
+  def assert_domains(domains)
+    expect(domains.count).to be > 0
+    assert_domain(domains.first)
+  end
+
+  def assert_domain(domain)
+    assert_time_fields(domain, %w(lastUpdatedDate creationDate))
+    assert_identifier_fields(domain, %w(ID lastUpdatedBy owner parentID templateID enterpriseID))
+    assert_string_fields(domain, %w(parentType name description))
+    assert_integer_fields(domain, %w(backHaulVNID serviceID customerID labelID ECMPCount domainID domainVLANID))
+    assert_boolean_property_fields(domain, %w(stretched globalRoutingEnabled leakingEnabled BGPEnabled))
+    assert_entity_scope(domain['entityScope'])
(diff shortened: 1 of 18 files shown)


gberginc:
@gasper-vrhovsek I've just submitted #19 that resolves the problem of failed Travis builds.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      vsd_client = ManageIQ::Providers::Nuage::NetworkManager::VsdClient.new("https://#{@hostname}:8443/nuage/api/v5_0", @userid, @password)
+
+      enterprises_resp = vsd_client.get_enterprises
+      domains_resp = vsd_client.get_domains
+      zones_resp = vsd_client.get_zones
+      subnets_resp = vsd_client.get_subnets
+      vports_resp = vsd_client.get_vports
+      vms_resp = vsd_client.get_vms
+      policy_groups_resp = vsd_client.get_policy_groups
+
```
miha-plesko: `end` the VCR.use_cassette block here just after you perform the API calls. The `expect` statements should be outside of that block.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      assert_common_parameters(vms_resp)
+
+      assert_object_not_empty(policy_groups_resp)
+      assert_json_object(policy_groups_resp)
+      assert_common_parameters(policy_groups_resp)
+    end
+  end
+
+  it "should fail on wrong password" do
+    VCR.use_cassette('manageiq/providers/nuage/network_manager/vsd_client/vsd_client_wrong_pass', :allow_unused_http_interactions => true, :record => :new_episodes) do
```
miha-plesko: `vsd_client/wrong_password` (shorten the path, and remove `vsd_client_` prefix since we're in `vsd_client` directory anyway). Also, remove the two arguments `:allow_unused_http_interactions => true, :record => :new_episodes`.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+        interaction.filter!(CGI.escape(@hostname), @hostname)
+        interaction.filter!(CGI.escape('NUAGE_NETWORK_HOST'), 'nuagenetworkhost')
+      end
+      c.filter_sensitive_data('NUAGE_NETWORK_AUTHORIZATION') { Base64.encode64("#{@userid}:#{@password}").chomp }
+    end
+  end
+
+  it "should return valid non empty json responses" do
+    VCR.use_cassette('manageiq/providers/nuage/network_manager/vsd_client/vsd_client', :allow_unused_http_interactions => true, :record => :new_episodes) do
+      vsd_client = ManageIQ::Providers::Nuage::NetworkManager::VsdClient.new("https://#{@hostname}:8443/nuage/api/v5_0", @userid, @password)
```
miha-plesko: Replace `ManageIQ::Providers::Nuage::NetworkManager::VsdClient` with `described_class`. You've specified what class are we describing in L2, hence this variable is now available here. Please do it in other tests below as well.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      domains_resp = vsd_client.get_domains
+      zones_resp = vsd_client.get_zones
+      subnets_resp = vsd_client.get_subnets
+      vports_resp = vsd_client.get_vports
+      vms_resp = vsd_client.get_vms
+      policy_groups_resp = vsd_client.get_policy_groups
+
+      assert_object_not_empty(enterprises_resp)
+      assert_json_object(enterprises_resp)
+      assert_common_parameters(enterprises_resp)
```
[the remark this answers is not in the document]
gberginc: Be careful about this one! Our current Nuage deployment is very dynamic, automatically changing the content of the database on the fly.

I agree that having such thing will be beneficial, however keep in mind that whenever the cassette will be recorded, we will probably have to adapt this as well.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      c.before_playback do |interaction|
+        interaction.filter!(CGI.escape(@hostname), @hostname)
+        interaction.filter!(CGI.escape('NUAGE_NETWORK_HOST'), 'nuagenetworkhost')
+      end
+      c.filter_sensitive_data('NUAGE_NETWORK_AUTHORIZATION') { Base64.encode64("#{@userid}:#{@password}").chomp }
+    end
+  end
+
+  it "should return valid non empty json responses" do
+    VCR.use_cassette('manageiq/providers/nuage/network_manager/vsd_client/vsd_client', :allow_unused_http_interactions => true, :record => :new_episodes) do
```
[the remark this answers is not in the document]
miha-plesko: EDIT: I've just realized that the first comment here (about `'vsd_client/vsd_client'` path) is inaccurate! But still, instead hardcoding the whole path, please do it like this:

```ruby
described_class.parent.name.underscore + 'vsd_client/vsd_client'
```

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    end
+  end
+
+  it "should fail on wrong username" do
+    VCR.use_cassette('manageiq/providers/nuage/network_manager/vsd_client/vsd_client_wrong_user', :allow_unused_http_interactions => true, :record => :new_episodes) do
+      expect do
+        ManageIQ::Providers::Nuage::NetworkManager::VsdClient.new("https://#{@hostname}:8443/nuage/api/v5_0", 'wrong_user', @password)
+      end.to raise_error(MiqException::MiqInvalidCredentialsError)
+    end
+  end
```
gberginc: If possible, add expectation on wrong host as well.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,109 @@
+require 'json'
```
gberginc: Remove `json` dependency as @miha-plesko suggested that we are not interested in JSON parsing at all.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
gberginc: @gasper-vrhovsek just to clarify, @miha-plesko means that you can simply replace this line with

```
vsd_client = described_class.new(...)
```

gasper-vrhovsek:
Thank you for commenting @miha-plesko and @gberginc . I'll prepare a patch ASAP fixing the issues you both highlighted.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      enterprises_resp = vsd_client.get_enterprises
+      domains_resp = vsd_client.get_domains
+      zones_resp = vsd_client.get_zones
+      subnets_resp = vsd_client.get_subnets
+      vports_resp = vsd_client.get_vports
+      vms_resp = vsd_client.get_vms
+      policy_groups_resp = vsd_client.get_policy_groups
+
+      assert_object_not_empty(enterprises_resp)
```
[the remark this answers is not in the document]
gasper-vrhovsek: Are we sure, we want to always expect 2 elements in this array? I know that currently the `get_policy_groups` returns 2 objects in an array, but will it always be so? @miha-plesko @gberginc please comment on this.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      assert_json_object(enterprises_resp)
```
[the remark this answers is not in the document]
gasper-vrhovsek: I see your point. If the json got successfully converted to a ruby object, it should be able to be converted back to JSON. I think we need to refactor `vsd_client.rb` (as you suggested) to be more testable.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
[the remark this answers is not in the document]
gberginc: You are recording the actual call to the service. At the time this was recorded, there were exactly 2 which @miha-plesko suggests to test. However this does not imply that `vsd_client.get_enterprises` must always return `enterprises.count == 2`.

That's one of the reasons why we are recording the calls so that we are able to mostly work in controlled environment. We could mock the backend, true, but VCR is here to help us.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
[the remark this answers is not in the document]
gasper-vrhovsek: `assert_common_parameters` was done according to the API documentation. In the section _4.1  REST API Conventions Overview_ these fields are described as _ common parameters of all objects because the all the objects available in the API have a
common structure_. Hence I wrote this test, which determines if the object returned is actually in compliance with the REST api convention.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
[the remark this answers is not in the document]
miha-plesko: As @gberginc said: the test and the VCR cassette can be bound together, I see no problem with that. Later when new VCR is recorded, the test must be updated accordingly.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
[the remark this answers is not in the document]
miha-plesko: Even if it wasn't convertible to JSON anymore, we're not converting it back to JSON anywhere in miq codebase.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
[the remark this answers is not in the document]
miha-plesko: Yup, I noticed that. But my understanding is that the test is not supposed to test compliance with rules, but to show developer at least one actual example response, so that developer can integrate further.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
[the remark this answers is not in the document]
gasper-vrhovsek: Ok, understood.

gberginc:
@gasper-vrhovsek #19 has been merged, so this one should be green once you submit the new patches. You can also close and reopen this PR to test it.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,596 @@
...
+        interaction.filter!(CGI.escape('NUAGE_NETWORK_HOST'), 'nuagenetworkhost')
+      end
+      c.filter_sensitive_data('NUAGE_NETWORK_AUTHORIZATION') { Base64.encode64("#{@userid}:#{@password}").chomp }
+    end
+  end
+
+  it "should return valid non empty json response for enterprises" do
+    enterprises = nil
+    VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/enterprises', :record => :new_episodes) do
+      vsd_client = described_class.new("https://#{@hostname}:8443/nuage/api/v5_0", @userid, @password)
```
gberginc: I suggest you move all `it` blocks into a higher level `context` block and make use of `before(:each)` to instantiate the `vsd_client`.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,596 @@
...
+    VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/domains', :record => :new_episodes) do
+      vsd_client = described_class.new("https://#{@hostname}:8443/nuage/api/v5_0", @userid, @password)
+      domains = vsd_client.get_domains
+    end
+
+    assert_object_not_empty(domains)
+    assert_domains_parameters(domains)
+  end
+
+  it "should return empty response for domains" do
```
gberginc: I am wondering - what does this do? You are calling invalid endpoint (`.../domains_empty`).

gasper-vrhovsek: No, i am using a VCR cassette named `domains_empty`. I have manually fixed the response in that cassette file to return UTF-8 encoded empty string, that is what the `vsd_client` expects in case the 'backend' has no data to return. This is how we cover the part of `vsd_client` methods which handle empty responses.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,990 @@
...
+      VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/login', :record => :new_episodes) do
+        @vsd_client = described_class.new("https://#{@hostname}:8443/nuage/api/v5_0", @userid, @password)
+      end
+    end
+
+    it "should return valid non empty response for get_enterprises" do
+      enterprises = nil
+      VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/enterprises', :record => :new_episodes) do
+        enterprises = @vsd_client.get_enterprises
+      end
```
miha-plesko: Do you think we could wrap these four repeating lines into a function like:

```ruby
def vcr_get(cassette, method, args: [])
	res = nil
	VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/' + cassette) do
	  res = @vsd_client.send(method, *args)
	end
end
```

Then here we would replace them with:

```ruby
enterprises = vcr_get('enterprises', :get_enterprises)
```

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,990 @@
...
+    it "should return empty response for policy_groups for domain" do
+      policy_groups = nil
+      VCR.use_cassette(described_class.parent.name.underscore + '/vsd_client/policy_groups_for_domain_empty', :record => :new_episodes) do
+        policy_groups = @vsd_client.get_policy_groups_for_domain(DOMAIN_ID)
+      end
+      expect(policy_groups).to be_nil
+    end
+  end
+
+  it "should fail on wrong password" do
```
miha-plesko: Not really neccessary, but I'd wrap these last three `it`-s into a
```ruby
context "when login fails" do ...
```
In RubyMine we will then be able to collapse them when needed 😄

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,990 @@
...
+    end
+  end
+
+  def assert_object_not_empty(object)
+    expect(object.length).to be > 0
+    expect(object.length).to be_truthy
+  end
+
+  def assert_enterprises(enterprises)
+    # test count
```
miha-plesko: From what I've seen, we're generally supposed to comment code only when really needed. I'd remove two comments here since it's obvious what we're doing.

miha-plesko: Also, I'm not sure that you need to check the count here again since you check it in each `it` explicitly by the means of `assert_object_not_empty`.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,990 @@
...
+      "enterpriseProfileID",
+      "receiveMultiCastListID",
+      "sendMultiCastListID",
+      "associatedGroupKeyEncryptionProfileID",
+      "associatedEnterpriseSecurityID",
+      "associatedKeyServerMonitorID",
+      "LDAPEnabled",
+      "LDAPAuthorizationEnabled",
+      "BGPEnabled",
+    )
```
miha-plesko: Wow, that hell of a list here, no wonder we're hitting 990 lines total 😄 If we wanted, we could write this list in a more compact way like this:

```ruby
expect(enterprise).to include(*%w(children parentType entityScope lastUpdatedBy lastUpdatedDate, ...))
```

But I leave it up to you to decide, it doesn't really bother me the way it is either.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,990 @@
...
+
+    # test for date values
+    assert_datetime(enterprise['lastUpdatedDate'])
+    assert_datetime(enterprise['creationDate'])
```
miha-plesko: I see we have many functions below to check parameter by parameter. May I suggest we group them into a "bulk" function:

```ruby
assert_datetimes(enterprise, %w(lastUpdatedDate creationDate))
assert_identifier_strings(enterprise, %w(ID lastUpdatedBy owner ...))
```

Having such "bulk" funcions implemented, number of lines would decrease notably.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,990 @@
...
+      "sendMultiCastListID",
+      "associatedGroupKeyEncryptionProfileID",
+      "associatedEnterpriseSecurityID",
+      "associatedKeyServerMonitorID",
+      "LDAPEnabled",
+      "LDAPAuthorizationEnabled",
+      "BGPEnabled",
+    )
+
+    # test for date values
```
miha-plesko: Comments here are not really needed IMO since we've named functions wisely 👍  so it's obvious what we're doing. Please remove them.

gasper-vrhovsek:
Hi @miha-plesko! Thanks for all the suggestion, i agree with all of them (or most of them at least, hehe). I'll implement as many changes as possible and submit an update asap.

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
gasper-vrhovsek: that counts number of objects in the list, the check in `assert_object_not_empty` checks if the object is empty (it shouldn't be)

gasper-vrhovsek:
Hi @miha-plesko and @gberginc. Would you take another look at the code to see if we have anything other to fix? 

Thanks

spec/models/manageiq/providers/nuage/vsd_client/vcr_specs/vsd_client_spec.rb
```diff
@@ -0,0 +1,642 @@
...
+      "public",
+      "templateID",
+      "associatedSharedNetworkResourceID",
+      "DHCPRelayStatus",
+      "proxyARP",
+      "multicast",
+      "associatedMulticastChannelMapID",
+      "DPI",
+      "useGlobalMAC"
+    )
```
miha-plesko: I just got wondering: do we even need to test for presence explicitly? I mean, we're testing each attribute in more details anyway so we could probably just enrich the "bulk" functions (`assert_time_fields`, `assert_identifier_fields`, ...) with presence check.

This way we wouldn't need to list all the attributes twice.

gasper-vrhovsek:
Hi @juliancheal,

we have reviewed this with @gberginc and @miha-plesko and think it is ready for your review.

@miq-bot remove_label wip
@miq-bot assign @juliancheal

juliancheal:
LGTM 👍

PR #19, opened 2017-10-18
Fix Qpid proton Travis build

gberginc:
Because qpid-proton needs to be built from source, additional system
packages are needed (installed with sudo rights). This caused that the
vendor bundle is created by the root user and resulted in Bundler not
being able to install other gems.

This patch now removes the use of sudo to the bare minimum, namely
installing the packages required to build qpid-proton from source.

### DIFF
--- a/bin/qpid_install.sh
+++ b/bin/qpid_install.sh
@@ -30,13 +30,13 @@ sed -i .bak -e 's/ 0/ 2/' proton-c/bindings/ruby/qpid_proton.gemspec.in
 # Configure the source of Qpid Proton.
 mkdir build
 cd build
-cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DSYSINSTALL_BINDINGS=OFF
+cmake .. -DCMAKE_INSTALL_PREFIX=$HOME/pkg -DSYSINSTALL_BINDINGS=OFF
 
 # Make system libraries and the gem.
 make all
 
 # Install system libraries
-sudo make install
+make install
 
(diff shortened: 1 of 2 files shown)


gberginc:
This patch temporarily addresses the problem with building of Qpid proton by not installing the the packages into system paths. We started with [this issue](https://travis-ci.org/ManageIQ/manageiq-providers-nuage/jobs/289302535) in Travis. The build is now passing so #16 should also work pass the installation process.

Once we will be including the `BUNDLE_WITH=qpid_proton`, I will update the build slightly.

@juliancheal @Fryguy Can you please take a look and merge?

bin/qpid_install.sh
```diff
@@ -30,13 +30,13 @@ sed -i .bak -e 's/ 0/ 2/' proton-c/bindings/ruby/qpid_proton.gemspec.in
...
 # Make system libraries and the gem.
 make all
 
 # Install system libraries
-sudo make install
+make install
 
 # Manually install the Ruby Gem. This will be installed inside $GEM_HOME directory.
-gem install proton-c/bindings/ruby/qpid_proton-0.18.0.gem
+gem install proton-c/bindings/ruby/qpid_proton-0.18.0.gem -- --with-qpid-proton-lib=$HOME/pkg/lib --with-qpid-proton-include=$HOME/pkg/include/
```
gberginc: I just want to point out that this is temporal patch to allow us to move some PRs forward. Once we merge the change in core that will install gem from main Gemfile, I will update the logic.

gberginc:
@miq-bot assign @juliancheal

PR #20, opened 2017-10-18
Implement targeted refresh for NetworkManager

miha-plesko:
With this commit we implement targeted refresh for NetworkManager. Following targets are supported, together with inferred co-targets:

```
TARGET           INFERS
-----------------------------------------------
NetworkGroup  -> cloud_subnets, security_groups
SecurityGroup -> network_group
CloudSubnet   -> network_group, security_groups
```

Read the table above like this: "When targeted refresh for NetworkGroup is triggered, also corresponding CloudSubnets and SecurityGroups get updated"... All three cases are tested in targeted_refresh_spec.rb.

To implement targeted refresh we had to slightly modify the existing parser so that we are able to share it for both graph refresh and targeted refresh.

@miq-bot assign @Ladas 

### DIFF
--- a/spec/models/manageiq/providers/nuage/network_manager/vcr_specs/targeted_refresh_spec.rb
+++ b/spec/models/manageiq/providers/nuage/network_manager/vcr_specs/targeted_refresh_spec.rb
@@ -1,0 +1,316 @@
+describe ManageIQ::Providers::Nuage::NetworkManager::Refresher do
+  TARGET_REFRESH_SETTINGS = [
+    {
+      :inventory_object_refresh => true,
+      :allow_targeted_refresh   => true,
+      :inventory_collections    => {
+        :saver_strategy => :default,
+      },
+    },
+    {
+      :inventory_object_refresh => true,
+      :allow_targeted_refresh   => true,
+      :inventory_collections    => {
+        :saver_strategy => :batch,
+        :use_ar_object  => true,
+      },
+    },
+    {
+      :inventory_object_refresh => true,
+      :allow_targeted_refresh   => true,
+      :inventory_collections    => {
+        :saver_strategy => :batch,
+        :use_ar_object  => false,
+      },
+    },
+    {
+      :inventory_object_saving_strategy => :recursive,
+      :inventory_object_refresh         => true,
+      :allow_targeted_refresh           => true,
+    }
+  ].freeze
+
+  before(:each) do
+    @ems = FactoryGirl.create(:ems_nuage_with_vcr_authentication, :port => 8443, :api_version => "v5_0", :security_protocol => "ssl-with-validation")
+  end
+
+  before(:each) do
+    userid   = Rails.application.secrets.nuage_network.try(:[], 'userid') || 'NUAGE_USER_ID'
+    password = Rails.application.secrets.nuage_network.try(:[], 'password') || 'NUAGE_PASSWORD'
+    hostname = @ems.hostname
+
+    # Ensure that VCR will obfuscate the basic auth
+    VCR.configure do |c|
+      # workaround for escaping host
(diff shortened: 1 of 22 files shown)


miha-plesko:
I think the Travis will currently fail since master is broken due to some unrelated problems, but I'm sending the implementation of targeted refresh anyway. @gberginc may I ask you for the first pass of review? The targeted collector is somewhat complicated since the Nuage API does not provide a better way to connect supported targets other than performing multiple API requests e.g. to get all CloudSubnets of a specific NetworkGroup. That's also the reason why I've implemented a simple cache to prevent those intermediate API calls from being repeted.

Looking forward to be hearing some comments!

gberginc:
@miha-plesko can you please close and reopen the PR to see if tests pass.

gberginc:
@miha-plesko now the master is already at 0.19... I will need yet another travis patch!

miha-plesko:
@gberginc my Travis is red and I like it! Because it means that new qpid got released and thus we will be able to finally provide event catcher for Nuage 😄

spec/models/manageiq/providers/nuage/network_manager/vcr_specs/targeted_refresh_spec.rb
```diff
@@ -0,0 +1,329 @@
...
+            )
+            test_targeted_refresh([cloud_subnet_target], 'cloud_subnet') do
+              assert_cloud_subnet_counts
+              assert_specific_cloud_subnet
+            end
+          end
+
+          it "will refresh cloud_subnet - object" do
+            cloud_subnet = FactoryGirl.create(:cloud_subnet_nuage, :ems_id => @ems.id, :ems_ref => cloud_subnet_ref1)
+            test_targeted_refresh([cloud_subnet], 'cloud_subnet') do
```
Ladas: right, so this one works, but MIQ will not allow this target through queue. The real integration test to prove it would be

example for containers, but it's all the same
```
target = ManagerRefresh::Target.new(
      :manager     => @ems,
      :association => :container_groups,
      :manager_ref => ems_ref,
      :options     => {
        :payload => notice['object'],
      }.to_json,
    )

    EmsRefresh.queue_refresh(target)
    expect(MiqQueue.where(:method_name => 'refresh').count).to eq 1
    refresh_job = MiqQueue.where(:method_name => 'refresh').first
    refresh_job.deliver
```

Ladas: But I suppose we can keep the current specs and add 1 more that will pass the target through queue

miha-plesko: I think the test in L71 above perfectly matches your description? 😉 
I mean, there is
```
it "will refresh cloud_subnet"
```
and
```
it "will refresh cloud_subnet - object"
```
already which covers what you describe here, I think.

miha-plesko: EDIT: ah, I thought you were referring to `ManagerRefresh::Target` vs providing the miq object directly. I see you actually meant to put to the queue, let me implement this as you suggest.

Ladas: yeah, so if you would put the ActiveRecord object into the Queue (which is the production workflow) you will see it fail, or ignore the target

miha-plesko:
Thanks for the comments @Ladas. I've updated the spec so that `test_targeted_refresh()` now always goes via `MiqQueue` to mock the production as closely as possible. Also Travis is now green so I'm happy to ask you for another pass of review.

PR #22, opened 2017-10-19
Extract vsd_client header into a function

miha-plesko:
With this commit we refactor client so that it doesn't produce code climate errors about repeated code anymore. Also we add a TODO since such hard-coded headers seem wrong.

@miq-bot add_label cleanup
@miq-bot assign @Ladas

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/vsd_client.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/vsd_client.rb
@@ -18,194 +18,86 @@ module ManageIQ::Providers
     end
 
     def get_enterprises
-      response = @rest_call.get(@server + '/enterprises')
-      if response.code == 200
-        if response.body == ''
-          _log.warn('No enterprises present')
-          return
-        end
-        return JSON.parse(response.body)
-      end
-      _log.error('Error in connection for server ' + @server.to_s + ' ' + response.code.to_s)
+      get_list('enterprises')
     end
 
     def get_enterprise(id)
-      response = @rest_call.get(@server + '/enterprises/' + id)
-      if response.code == 200
-        if response.body == ''
-          _log.warn("Enterprise with ID not found: #{id}")
-          return
-        end
-        return JSON.parse(response.body).first
-      end
-      _log.error('Error in connection for server ' + @server.to_s + ' ' + response.code.to_s)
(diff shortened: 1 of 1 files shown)


miha-plesko:
@Ladas this is a followup that you asked me for earlier. I see there are some new code climate issues now, but please note that refactoring of the `vsd_client.rb` is out of our scope at the moment. That being said, it's obvious that the refactored is really needed since it currently handles API errors quite poorly...

app/models/manageiq/providers/nuage/network_manager/vsd_client.rb
```diff
@@ -78,8 +78,7 @@ def get_domains_for_enterprise(enterprise_id)
     end
 
     def get_zones
-      @rest_call.append_headers("X-Nuage-FilterType", "predicate")
-      @rest_call.append_headers("X-Nuage-Filter", "name ISNOT 'BackHaulZone'")
+      exclude_name('BackHaulZone')
```
Ladas: seems like the whole function can be extracted to something like

get_list('zones')
get_list('subnets') 
etc.

since the collection name seems to be the only difference

gberginc: Although I agree we can have a `get_list` that will return the collection, I would nevertheless prefer that we keep `get_zones` and `get_subnets` that in turn call `get_list`. Otherwise, people need to know exactly what to pass as arguments.

@Ladas would this be ok for you?

Similarly we could update `get_zone` and `get_subnet` that call just `get_object(type, id)`.

Ladas: yes, it's just about abstraction the similar code to one place, so 

```
def get_zones
   get_list('zones')
end
```

gberginc: @miha-plesko Since you started this, can you please cleanup the code as @Ladas suggested? It should be a fairly trivial change and will also make codeclimate happy.

miha-plesko:
@Ladas green. Apologies for such a delay.

BTW, we're working on [VCR unit tests](https://github.com/ManageIQ/manageiq-providers-nuage/pull/16) for vsd_client as well, so that things will get controlled.

PR #23, opened 2017-10-19
Remove get_vports method from vsd_client.rb

gasper-vrhovsek:
Cause for removal is that calls to vports on api levels 4.0 and 5.0 fail with
404 error code. Besides this method is not used anywhere and is redundant.


gberginc:
I am fine with the removal as well!

@juliancheal can we merge?

miha-plesko:
@gasper-vrhovsek this API call was removed as part of https://github.com/ManageIQ/manageiq-providers-nuage/pull/22, so I think we can just close the PR.

PR #25, opened 2017-10-24
Install qpid_proton gem from core Gemfile

gberginc:
Previously, the gem was installed directly from the source. However,
since the `qpid_proton` gem has been incorporated as optional in the
core `Gemfile` it is now also possible to use that gem directly by
setting the environment variable `BUNDLE_WITH=qpid_proton`.

This patch updates Travis to use the gem from `Gemfile`.

### DIFF
--- a/bin/qpid_install.sh
+++ b/bin/qpid_install.sh
@@ -1,17 +1,5 @@
 #!/bin/bash
 
-# Ensure correct Ruby environment is used.
-. ~/.rvm/scripts/rvm
-rvm use $TRAVIS_RUBY_VERSION
-
-# We have to extract the major and minor ruby versions so that we can set proper GEM_HOME
-ruby_version=`echo $TRAVIS_RUBY_VERSION | egrep -o '[[:digit:]]\.[[:digit:]]'`
-
-# As we are installing the Gem built from source, we have to set the proper
-# GEM_HOME. This will ensure that the gem will be installed in a place where it
-# will be picked by bundler.
-export GEM_HOME=/home/travis/build/ManageIQ/manageiq-providers-nuage/vendor/bundle/ruby/$ruby_version.0
-
 # Install the dev dependencies for building Qpid proton system library.
 sudo apt-get install -y gcc cmake cmake-curses-gui uuid-dev
 sudo apt-get install -y libssl-dev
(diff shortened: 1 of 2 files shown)


gberginc:
@miq-bot remove_label wip

gberginc:
@juliancheal @Fryguy @Ladas this now properly installs the `qpid_proton` gem using the specification in the core `Gemfile`. Consequently, the build script is greatly simplified. I have tested this patch with a simple `require 'qpid_proton'` inside one of the existing specs to ensure that the gem is available. The result was green [here](https://travis-ci.org/ManageIQ/manageiq-providers-nuage/builds/291899868) -- I then removed that `require` because it does not fit into this PR, but will be added in following PRs.

I would kindly ask for a quick merge so that I can update #9.

gberginc:
I've cherrypicked this PR onto #9 to check that it works and it's all green now 😄.

@miq-bot assign @juliancheal

juliancheal:
@Fryguy I just wanna check with you this is OK, before i merge.

bin/qpid_install.sh
```diff
@@ -23,20 +11,13 @@ cd $HOME/build
...
 
-# There is a strange dependency on JSON that we need to change to version of
-# at least 2; otherwise there will be a conflict.
-sed -i .bak -e 's/ 0/ 2/' proton-c/bindings/ruby/qpid_proton.gemspec.in
-
 # Configure the source of Qpid Proton.
 mkdir build
 cd build
-cmake .. -DCMAKE_INSTALL_PREFIX=$HOME/pkg -DSYSINSTALL_BINDINGS=OFF
+cmake .. -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_BINDINGS= -DSYSINSTALL_BINDINGS=OFF
```
Fryguy: Once you remove bindings (yay, btw!), you can also remove swig

Fryguy:
I'm going to merge, to let @gberginc continue.  @gberginc can you remove swig in a separate PR?

bin/qpid_install.sh
gberginc: Yes, I will create a separat PR for that. Thanks!

PR #28, opened 2017-10-25
Connect events to targeted refresh

miha-plesko:
With this commit we teach our NetworkManager how to map obtained events to appropriate targets. At the moment only events concerning the three supported targets actually trigger something:

```
- network_groups
- cloud_subnets
- security_groups
```

while all other events are simply skipped.

Depends on: https://github.com/ManageIQ/manageiq-providers-nuage/pull/9

@miq-bot assign @Ladas 

/cc @gberginc


miha-plesko:
This PR will fail until EventCatcher isn't merged but I'd kindly ask you @Ladas to take a look at it anyway (it's a very short PR).

PR #29, opened 2017-10-26
Use qpid_proton from rubygems

gberginc:
Now that `qpid_proton` gem has been officially released on RubyGems, we
can remove external dependency on the temporary GitHub repository that
hosted Ruby bindings for Qpid. This requires addition of the
`qpid_proton` gem in the gemspec.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ env:
   - RUBY_GC_HEAP_GROWTH_MAX_SLOTS=300000
   - RUBY_GC_HEAP_INIT_SLOTS=600000
   - RUBY_GC_HEAP_GROWTH_FACTOR=1.25
-  - BUNDLE_WITH=qpid_proton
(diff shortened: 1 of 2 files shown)


gberginc:
@miq-bot assign @juliancheal 

This is part one of the update. The other will require removal of the `qpid_proton` from the core.

chargio:
It is great that they have released the final gem!

pdellaert:
Superb news, indeed!

juliancheal:
Linked to https://github.com/ManageIQ/manageiq/pull/16307

juliancheal:
The problem is, the `qpid_proton` has `:optional => true` so by moving the gem to the `gem spec` it fails to install unless a developer has the dependencies that `qpid_proton` requires.

Not sure how we can keep the spirit of `:optional => true` here.

PR #31, opened 2017-10-27
[WIP] Expose auth_url in Nuage provider as a class method

gasper-vrhovsek:
Because of changes in the way the validation requests are executed
in manageiq-ui-classic we have to expose the auth_url method as a
class method. This change is required for:
https://github.com/ManageIQ/manageiq-ui-classic/pull/2532

@miq-bot add_label networks

PR #32, opened 2017-10-27
Update raw_connect to simplify validation from UI

gberginc:
The UI repo has changed the way endpoint validation occurs, causing a
problem withinin the angular controller when creating the arguments for
`raw_connect`. Previously, the `auth_url` was required as an argument
which mandated the need to expose the `auth_url` from the Nuage manager
so that it could be used in the UI controller.

Instead, this patch replaces `auth_url` with the specific arguments that
are needed to build the `auth_url` so that the UI can directly pass
those values into the manager.

### DIFF
--- a/app/models/manageiq/providers/nuage/manager_mixin.rb
+++ b/app/models/manageiq/providers/nuage/manager_mixin.rb
@@ -2,24 +2,37 @@ module ManageIQ::Providers::Nuage::ManagerMixin
   extend ActiveSupport::Concern
 
   module ClassMethods
-    def raw_connect(auth_url, username, password)
-      ManageIQ::Providers::Nuage::NetworkManager::VsdClient.new(auth_url, username, password)
+    def raw_connect(username, password, endpoint_opts)
+      protocol    = endpoint_opts[:protocol].strip if endpoint_opts[:protocol]
+      hostname    = endpoint_opts[:hostname].strip
+      api_port    = endpoint_opts[:api_port]
+      # In case API port is represented as a string, ensure it has no whitespaces.
+      api_port.strip! if api_port.kind_of?(String)
+      api_version = endpoint_opts[:api_version].strip
+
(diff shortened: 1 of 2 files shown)


gberginc:
@miq-bot assign @blomquisg 

The purpose of this PR is to offer an alternative to #31. Even though #31 helps the UI to validate the endpoint, I would rather use the approach of most other providers that simply get credentials and endpoint configuration. One of the additional benefits is that the arguments are now stripped (as other providers) because previously we were actually able to break the provider when whitespaces were present in the params.

@blomquisg If you agree with this one, we will close #31 and then also update the UI part.

/cc @gasper-vrhovsek

app/models/manageiq/providers/nuage/manager_mixin.rb
```diff
@@ -2,24 +2,37 @@ module ManageIQ::Providers::Nuage::ManagerMixin
...
+      api_version = endpoint_opts[:api_version].strip
+
+      url = auth_url(protocol, hostname, api_port, api_version)
+      _log.info("Connecting to Nuage VSD with url #{url}")
+      ManageIQ::Providers::Nuage::NetworkManager::VsdClient.new(url, username, password)
+    end
+
+    def auth_url(protocol, server, port, version)
+      scheme = protocol == "ssl-with-validation" ? "https" : "http"
+      "#{scheme}://#{server}:#{port}/nuage/api/#{version}"
```
Fryguy: This should use a URI builder instead (though I see this is a copy, so it can be done in a separate PR).  The main reason is that for IPv6, if an IPv6 server address is given, this code won't actually work...prefer

```ruby
URI::Generic.build(:scheme => scheme, :host => server, :port => port, :path => "/nuage/api/#{version}")
```

gberginc:
@blomquisg @Fryguy I have addressed the comment made by @Fryguy regarding the use of `URI::Generic.build`. I've added a spec to make sure URLs did not change.

PR #33, opened 2017-10-30
Enable Nuage network manager

gberginc:
Prior to this patch, the UI could not show the option to create new
network managers because these are typically created along with cloud
manager. Nuage is a dedicated SDN requiring ability to add the provider
manually.

The UI depends on the `product.nuage` setting to enable the network
providers which this patch adds directly into the Nuage.

@miq-bot assign @blomquisg 

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -14,6 +14,8 @@
   :nuage_network:
     :inventory_object_refresh: true
     :allow_targeted_refresh: true
+:product:
+  :nuage: true
 :workers:
   :worker_base:
     :event_catcher:


djberg96:
@blomquisg @gberginc IMO this PR should be reverted because it's causing an issue where it will always enable the UI edit/new menu. Instead we should rely on the `SupportsFeatureMixin` in conjunction with a modification to the UI code.

https://github.com/ManageIQ/manageiq-ui-classic/pull/3976

https://bugzilla.redhat.com/show_bug.cgi?id=1577888

miha-plesko:
Will test tomorrow but I think [the PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/3976) will prevent user from adding the first Nuage manager. I'm thinking if we should instead check for `config/permissions.yaml` on the UI (where we decide to show button or not) if it contains entry for Nuage so that user would only have to enable Nuage there?

djberg96:
@miha-plesko Pretty sure you're correct. I'm not familiar with config/permissions.yaml, but it sounds promising.

PR #34, opened 2017-10-30
Fix Nuage api_version

gberginc:
Previously it was possible to pass an empty API version causing the
provider to fail. This patch updates the provider to set the default API
version to the latest version (v5_0), which is also the first one going
to be supported by this provider. Specs are updated to address this
change.

@miq-bot add_label bug


gberginc:
@miq-bot assign @blomquisg

gberginc:
Travis progressing, so it should get green soon.

PR #41, opened 2017-11-13
[WIP] Batch collecting of Nuage events with qpid sleeping

miha-plesko:
With this commit we force periodic interrupts to the EventCatcher
thread because the gem for accessing the AMQP queue seems to consume
all EventCatcher thread resources otherwise. By default we now interrupt
the qpid gem every 15 seconds and let it sleep for 15 seconds. While
the qpid thread is sleeping, the EventProcessing thread has enough
resources to quickly process events that were captured so far.

It is not clear at the moment why resources are not distributed equally
among the threads - could be a leak in amqp_proton or in MIQ. Below please
find description of old and new event catcher behavior:

--- OLD BEHAVIOR ---
Events were reported immediately by qpid_proton gem, but they didn't get
processed by ManageIQ::Providers::Nuage::NetworkManager::EventParser for quite
a while then. Since events were being continously collected in the queue and
EventParser didn't have resources to process them the delay was increasing.
Soon after MIQ boot the delay between receiveing event and processing event
was as huge as 10 min.

--- NEW BEHAVIOR ---
Events are reported while the qpid thread is running and they are all processed
instantly when the qpid thread is put to sleep. So each event gets processed
in about 15 seconds, which is acceptable.

@miq-bot assign @juliancheal 

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
@@ -23,16 +23,22 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream
     require 'qpid_proton'
 
     @options = options
+    @thread  = nil
+    @batch   = Queue.new # thread-safe implementation of Array
   end
 
-  def start(&message_handler_block)
+  def start_batch
     _log.debug("#{self.class.log_prefix} Opening amqp connection using options #{@options}")
-    @options[:message_handler_block] = message_handler_block if message_handler_block
-    connection.run
+    @options[:message_handler_block] = ->(event) { @batch << event }
+    @thread = qpid_thread
+    while @thread.alive? || !@batch.empty?
+      sleep_poll_normal
+      yield current_batch unless @batch.empty?
+    end
   end
 
   def stop
-    @handler.stop
+    @thread.exit if @thread && @thread.alive?
   end
 
   def connection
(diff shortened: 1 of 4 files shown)


miha-plesko:
@gasper-vrhovsek @gberginc I kindly ask for a first pass of review.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -42,4 +48,39 @@ def connection
...
+  #
+  # A simple thread `Thread.new { connection.run }` should normally be returned here.
+  # But doing so we measure a huge performance drop: events are fetched from AMQP immediately,
+  # but their processing is delayed for ~10min, which is horrible. It seems to be a problem
+  # with qpid somehow eating resources of the main thread. A quick workaround for now is to
+  # periodically pause qpid thread for a few seconds to let the main process access the resources.
+  #
+  def qpid_thread
+    t_interval = self.class.parent.worker_settings[:qpid_rest_interval]
+    t_sleep = self.class.parent.worker_settings[:qpid_rest_time]
```
gasper-vrhovsek: Maybe make these two arguments? You can then configure the interval and sleep for the qpid thread from outside. I think it would be more flexible and easier to test.

gberginc:
@miha-plesko My problem [here](https://github.com/ManageIQ/manageiq-providers-nuage/pull/41/files#diff-c15151955f0be6d8a5f131455906f2adR82) is that we loose the events during this `sleep`. I understand this solves the problem of getting the events into the system faster, however loosing them is not what we want.

We need to find a better way of doing this.

@miq-bot add_label wip
@miq-bot assign @gberginc

gberginc:
https://github.com/ManageIQ/manageiq-providers-nuage/issues/47 describes further details of the problem. Issue also mentions an alternative solution which needs to be discussed.

Ladas:
@miha-plesko @gberginc right, so how ruby green threads work, in switch context only if there is no work for some time, e.g. waiting for IO operation, DB, or sleep in this case. 

Btw. how are we loosing data with the sleep? AFAIK every AMQP deals with this by having the messages in the queue, until you read it (at least up to some size)? Or what is the exact setup here?

cc @agrare

gberginc:
@Ladas have you perhaps seen https://github.com/ManageIQ/manageiq-providers-nuage/issues/47? Does it make sense to simply include that `sleep_poll_normal`?

We can also add batched processing if that makes more sense, like in this PR of @miha-plesko, however I don't think that's the main cause for the delays.

Regarding my comment on the [`sleep`](https://github.com/ManageIQ/manageiq-providers-nuage/pull/41/files#diff-c15151955f0be6d8a5f131455906f2adR82) and loosing events. This `sleep` occurs while the connection to AMQP server is closed. Based on our tests, I believe that any message that is received before the connection is reestablished is lost. We have a small demo app that does what this event catcher does and whenever we start it we start fresh.

miha-plesko:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-nuage/pull/58

PR #45, opened 2017-11-20
Fix protocol selection when adding a new nuage provider

gasper-vrhovsek:
This fixes endpoint url generation from form input values. Before, if SSL was selected, the url began
with 'http'. Now when either 'SSL' or 'SSL with validation' are selected, the url begins with 'https'.

@miq-bot add_label bug

/cc @gberginc 
/cc @miha-plesko

### DIFF
--- a/app/models/manageiq/providers/nuage/manager_mixin.rb
+++ b/app/models/manageiq/providers/nuage/manager_mixin.rb
@@ -17,7 +17,7 @@ module ManageIQ::Providers::Nuage::ManagerMixin
     end
 
     def auth_url(protocol, server, port, version)
-      scheme = protocol == "ssl-with-validation" ? "https" : "http"
+      scheme = %w(ssl ssl-with-validation).include?(protocol) ? "https" : "http"
       URI::Generic.build(:scheme => scheme, :host => server, :port => port, :path => "/nuage/api/#{version}").to_s
     end
   end
--- a/spec/models/manageiq/providers/nuage/network_manager_spec.rb
+++ b/spec/models/manageiq/providers/nuage/network_manager_spec.rb
@@ -17,6 +17,12 @@ describe ManageIQ::Providers::Nuage::NetworkManager do
       @ems.connect
     end
 
+    it 'connects over secure channel without validation' do
+      @ems.security_protocol = 'ssl'
+      expect(ManageIQ::Providers::Nuage::NetworkManager::VsdClient).to receive(:new).with("https://host:8000/nuage/api/v5_0", "testuser", "secret")
+      @ems.connect
+    end
+
     it 'connects over secure channel' do
       @ems.security_protocol = 'ssl-with-validation'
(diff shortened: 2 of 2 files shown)


gberginc:
@gasper-vrhovsek please fix rubocop issues.

@miq-bot add_label gaprindashvili/yes

app/models/manageiq/providers/nuage/manager_mixin.rb
```diff
@@ -17,7 +17,7 @@ def raw_connect(username, password, endpoint_opts)
     end
 
     def auth_url(protocol, server, port, version)
-      scheme = protocol == "ssl-with-validation" ? "https" : "http"
+      scheme = !protocol || (protocol == "non-ssl") ? "http" : "https"
```
miha-plesko: I find this `!protocol` part confusing*, perhaps we would be better off with:
```
scheme = protocol.to_s.start_with? "ssl" ? "https" : "http"
```

*\*Too much Java-like. I mean, we're in ruby, we shouldn't check for `null` all over the code :wink:*

gasper-vrhovsek: You mean `nil` ? :)

gasper-vrhovsek:
I have reverted to the `!protocol` solution, as otherwise this test fails:

https://github.com/ManageIQ/manageiq-providers-nuage/blob/master/spec/models/manageiq/providers/nuage/network_manager_spec.rb#L144

app/models/manageiq/providers/nuage/manager_mixin.rb
```diff
@@ -17,7 +17,7 @@ def raw_connect(username, password, endpoint_opts)
     end
 
     def auth_url(protocol, server, port, version)
-      scheme = protocol == "ssl-with-validation" ? "https" : "http"
+      scheme = (!protocol || protocol == "non-ssl") ? "http" : "https"
```
bronaghs: @gasper-vrhovsek 
This is difficult to read. Can you describe the logic here so I can make sure I understand it correctly?

gasper-vrhovsek: Hi @bronaghs, sure i can describe:

If protocol is not provided or if 'non-ssl' is selected, then 'http' value is assigned. Otherwise 'https' is assigned.

My background is mostly java, so I am still getting used to Ruby. I'll search for a more readable solution.

gasper-vrhovsek: @bronaghs and @miha-plesko I have refactored the scheme selecting logic. Hope it's better and more readable now.

bronaghs:
Thanks @gasper-vrhovsek for reworking it 👍

simaishi:

```
$ git log -1
commit 935da2208632d5ae3f78106c0a4daf96213686dc
Date:   Wed Dec 13 09:57:10 2017 +0000

    Merge pull request #45 from gasper-vrhovsek/feature/fix_protocol_selection
    
    Fix protocol selection when adding a new nuage provider
    (cherry picked from commit 3887e05bdda08c68c784450abe524e99a2f1bbe0)
```

PR #48, opened 2017-11-29
[WIP] Fix delay of event processing by terminating qpid connection

gasper-vrhovsek:
This fix will terminate event connection every time when there are
no queued events in the event receiver and the receiver is drained.

@miq-bot add_label wip

/cc @miha-plesko @gberginc


gasper-vrhovsek:
@gberginc this is an alternative solution to https://github.com/ManageIQ/manageiq-providers-nuage/pull/41 from @miha-plesko. With this there was no delays between receiving events and them being processed.

miha-plesko:
@gasper-vrhovsek please close this PR in favor of https://github.com/ManageIQ/manageiq-providers-nuage/pull/58

PR #51, opened 2017-12-11
Url encode Nuage credentials

gasper-vrhovsek:
Encode username and password to avoid errors when using special characters such as `@` inside username and/or password. Because the amqp endpoint is generated as:
`[userid]:[password]@[amqp.endpoint.url]`

When, for instance, password contained `@` it got generated as something like:

`userid:p@ssword@amqp.endpoint.url`

which caused the validate call to get stuck in an infinite loop.

@miq-bot add_label bug

/cc @gberginc @miha-plesko

### DIFF
--- a/spec/models/manageiq/providers/nuage/network_manager_spec.rb
+++ b/spec/models/manageiq/providers/nuage/network_manager_spec.rb
@@ -73,6 +73,21 @@ describe ManageIQ::Providers::Nuage::NetworkManager do
         expect { @ems.verify_credentials(:amqp) }.to raise_error(StandardError)
       end
     end
+
+    context 'credentials encoding' do
+      before do
+        creds = {}
+        creds[:amqp] = {:userid => "amqp_user@!$&", :password => "amqp_password@!$&"}
+        @ems.endpoints << Endpoint.create(:role => 'amqp', :hostname => 'amqp_hostname', :port => '5672')
+        @ems.update_authentication(creds, :save => false)
+      end
+
+      it 'encodes username and password' do
+        opts = @ems.event_monitor_options
+        hostname = opts[:urls].first
+        expect(hostname).to eq('amqp_user%40%21%24%26:amqp_password%40%21%24%26@amqp_hostname:5672')
+      end
+    end
   end
 
   context 'translate_exception' do
--- a/app/models/manageiq/providers/nuage/manager_mixin.rb
+++ b/app/models/manageiq/providers/nuage/manager_mixin.rb
@@ -95,7 +95,7 @@ module ManageIQ::Providers::Nuage::ManagerMixin
 
     amqp_endpoints.map do |e|
       url = "#{e.hostname}:#{e.port}"
-      url = "#{amqp_auth.userid}:#{amqp_auth.password}@#{url}" if amqp_auth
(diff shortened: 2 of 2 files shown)


gasper-vrhovsek:
Hi @juliancheal 

This is just a minor bugfix regadring url encoded parameters when building the AMQP endpoint url.

@miq-bot assign @juliancheal

chargio:
Is there no impact on tests?

juliancheal:
Are there any existing tests for this? If not would be good to add some.

juliancheal:
Also would like to see if this affects other providers. This could be something we do for all credentials. I'll look into it.

gasper-vrhovsek:
All current tests pass. I agree, it would be good to have some. As for impacting other providers, i am not sure if this mixin is used elsewhere?

Before this fix we actually used an username which contained '%40' instead of '@'.

juliancheal:
@gasper-vrhovsek you're right no one else uses this Mixin, but I it occurred to me other providers which use `username` & `password` could have similar issues. Mainly a note for myself 😃

gasper-vrhovsek:
@juliancheal no problem :) Better safe than sorry. I'll look into possible ways of testing this. But probably the correct way would be qpid-proton raising an exception upon receiving a malformed url.

gasper-vrhovsek:
@gberginc & @miha-plesko maybe we should implement validation of username and password fields?

miha-plesko:
If you're referring to frontend validation (angular) then I think it's not really needed because we test on the backend. I mean, the UX would increase, but I'm not sure if it's worth it.

That being said: in angular you can probably simply specify the regex directly in haml, right? If yes, then it would be maybe even worth it.

gasper-vrhovsek:
Well, I am thinking if validation will solve anything. We do not forbid non url safe characters in username/password. We just need to encode them before we pass them to qpid-proton.

gasper-vrhovsek:
Rspec tests for encoding added.

spec/models/manageiq/providers/nuage/network_manager_spec.rb
```diff
@@ -73,6 +73,22 @@
...
+    context 'credentials encoding' do
+      before do
+        creds = {}
+        creds[:amqp] = {:userid => "amqp_user@!$&", :password => "amqp_password@!$&"}
+        @ems.endpoints << Endpoint.create(:role => 'amqp', :hostname => 'amqp_hostname', :port => '5672')
+        @ems.update_authentication(creds, :save => false)
+      end
+
+      it 'encodes username and password' do
+        # @ems.send(:amqp_urls)
```
bronaghs: @gasper-vrhovsek - Can you delete this commented out line if it isn't needed.

gasper-vrhovsek: @bronaghs I apologize for the late fix, the commented line is now removed.

simaishi:
@gasper-vrhovsek Backporting to Fine branch requires a BZ, please add BZ link.

simaishi:

```
$ git log -1
commit 647b6629f063cb18f5e9c91245ddc989d7653291
Date:   Thu Jan 4 13:50:23 2018 +0000

    Merge pull request #51 from gasper-vrhovsek/feature/encode_credentials
    
    Url encode Nuage credentials
    (cherry picked from commit 39f56e3f84365ee55198a401c04b9c47605fd09a)
```

Issue #47, opened 2017-11-29
Long delays processing Nuage events


gberginc:
We are experiencing long delays when listening for Nuage events in this provider. What's worse is that delays are increasing over time. Check this dump from the database (query used was `select id, event_type, timestamp, created_on, EXTRACT(EPOCH FROM (created_on - timestamp)) from event_streams where source = 'NUAGE' order by id desc`):

<img width="770" alt="screen shot 2017-11-29 at 08 01 50" src="https://user-images.githubusercontent.com/1437960/33366900-fc83f106-d4ec-11e7-85b0-3ce87b2c1b9e.png">

Top rows indicate 10 minute delays.

Adding `sleep_poll_normal` after [this line](https://github.com/ManageIQ/manageiq-providers-nuage/blob/master/app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb#L19) reduces this dramatically:
<img width="760" alt="screen shot 2017-11-29 at 08 01 34" src="https://user-images.githubusercontent.com/1437960/33366996-468d8fe6-d4ed-11e7-8d8b-ed527d89813f.png">

Some events still take 20 or 30 seconds (something we need to investigate further), but in general this looks much better.

@bronaghs @Ladas @agrare @Fryguy Is there any other way of overcoming these issues? We based this event catcher on [Amazon](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/cloud_manager/event_catcher/runner.rb#L12), which also does not use `sleep_poll_normal`, but I don't know the details of the stream implementation which could be the main difference.

@miq-bot assign @gberginc
/cc @miha-plesko, @gasper-vrhovsek

gberginc:
After a [comment](https://github.com/ManageIQ/manageiq-providers-nuage/pull/41#issuecomment-347804909) made by @Ladas about the green threads and after a discussion with @justinc1 and @miha-plesko, the problem could be related to the fact that qpid gem is only a binding for a system package which means that the "wait for IO" does not occur directly in Ruby, so perhaps context switches are not handled properly.

@justinc1 Was this my correct interpretation?

Ladas:
@gberginc from my personal experience, e.g. doing lots and lots of very quick DB queries never switches context. So the sleep_poll_normal is the way to make sure the context is switched.

justinc1:
[...]

PR #52, opened 2017-12-19
Improve the performance of event catcher

gberginc:
Nuage event catcher relies on Qpid Proton to get the events from the
AMQP queue. Due to synchronization issues, this resulted in massive
delays when processing events.

This patch tries to overcome this by adding a short sleep cycle to allow
the stream (Qpid) and event catcher to work together on the events.

It also resolves an issue with the fact the Nuage events are sometimes
sent with the same timestamp for certain event types, even though the
actual payload differs. As there is no ID that could be used as an EMS
reference, we alter the timestamp slightly based on the current time.

Addresses #47.
Replaces #41 and #48. Main reason for this is that I don't want to close the connection as we are loosing events! I've also attempted to do batch processing but it's even worse. Thus, for the first improvement I would like to see this merged as it improves the performance in a way that events are published into the DB.

@miq-bot add_label events, bug

@miq-bot assign @juliancheal 

/cc @bronaghs I'd also like your comment on this one, if possible.

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
@@ -17,6 +17,7 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Runner < ManageI
     event_monitor_handle.start do |event|
       event_monitor_running
       @queue.enq(event)
+      sleep_poll_normal
     end
   ensure
     stop_event_monitor
--- a/app/models/manageiq/providers/nuage/network_manager/event_parser.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_parser.rb
@@ -1,10 +1,14 @@
 module ManageIQ::Providers::Nuage::NetworkManager::EventParser
   def self.event_to_hash(event, ems_id)
     event_type = "#{event['entityType']}_#{event['type'].downcase}"
+    # Nuage pushes several events with the same eventReceivedTime causing
+    # EmsEvent to ignore some of them because of the same primary key. Here we
+    # just add several millisecconds to the event.
+    time = event["eventReceivedTime"] + Time.now.usec / 1000
     {
       :source     => "NUAGE",
       :event_type => event_type,
-      :timestamp  => DateTime.strptime((event["eventReceivedTime"] / 1000).to_s, '%s').to_s,
+      :timestamp  => DateTime.strptime(time.to_s, '%Q').strftime('%F %T.%6N'),
       :message    => event_type,
       :vm_ems_ref => nil,
       :full_data  => event.to_hash,
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -22,3 +22,4 @@
       :event_catcher_nuage_network:
         :topics:
           - topic/CNAMessages
+        :poll: 1.seconds


gberginc:
Hey @miha-plesko and @gasper-vrhovsek, all that rumbling and then this PR... What do you guys think? Any idea how to overcome that problem with the same timestamps?

Basically, what happens is that one event, e.g. `gwstate`, is triggered multiple times with different payloads but the same timestamp. Because of [this](https://github.com/manageiq/manageiq/blob/master/app/models/ems_event.rb#L243-L249) only the first occurrence of this burst gets added into the DB.

config/settings.yml
```diff
@@ -22,3 +22,4 @@
       :event_catcher_nuage_network:
         :topics:
           - topic/CNAMessages
+        :poll: 1.seconds
```
miha-plesko: Please add empty line at the end of the file

app/models/manageiq/providers/nuage/network_manager/event_parser.rb
```diff
@@ -1,10 +1,14 @@
...
     event_type = "#{event['entityType']}_#{event['type'].downcase}"
+    # Nuage pushes several events with the same eventReceivedTime causing
+    # EmsEvent to ignore some of them because of the same primary key. Here we
+    # just add several millisecconds to the event.
+    time = event["eventReceivedTime"] + Time.now.usec / 1000
     {
       :source     => "NUAGE",
       :event_type => event_type,
-      :timestamp  => DateTime.strptime((event["eventReceivedTime"] / 1000).to_s, '%s').to_s,
+      :timestamp  => DateTime.strptime(time.to_s, '%Q').strftime('%F %T.%6N'),
```
miha-plesko: Please add [test case](https://github.com/ManageIQ/manageiq-providers-nuage/blob/master/spec/models/manageiq/providers/nuage/network_manager/event_target_parser_spec.rb) to test the final value that appears in `:timestamp` field.

app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
```diff
@@ -17,6 +17,7 @@ def monitor_events
     event_monitor_handle.start do |event|
       event_monitor_running
       @queue.enq(event)
+      sleep_poll_normal
```
miha-plesko: Didn't we test this long ago already and it didn't help much? I'm surprised why it would help now. What's the throughput now?

config/settings.yml
gberginc: Hmmm, not sure. Looking at the changes, I don't think I actually removed it. I checkout `master` and there also is no newline (nor is there any in OpenStack provider).

app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
gberginc: There were so many things that I have removed for this PR that it really hurt. The most important one was the batched processing which I thought will help the most, but there are actually plenty of sync issues. In the end, this is what was needed. I know, my LOC metric will not be good this time 😄.

config/settings.yml
miha-plesko: Wanna bet for a burger? 😉 

There is empty trailing line both in Amazon and OpenStack provider's settings.yaml, only GitHub won't render it. Instead, if there is no empty line at the end of file, GitHub would render the ugly red sign, like in this case.

capture

app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
bronaghs: @gberginc - How will a sleep address the delay?

config/settings.yml
bronaghs: Will this `:poll` setting be read or will the one in the manageiq repo be read instead? See  [here](https://github.com/ManageIQ/manageiq/blob/master/config/settings.yml#L1070). The reason i ask is that the `sleep_poll_normal` looks for a `workers` section of the settings. 
See:
https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_worker/runner.rb#L448..#L450
https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_worker/runner.rb#L302
https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_worker.rb#L221..#L223

app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
gberginc: Problem is that without this delay, the processing of events does not start. With this sleep, we allow the other thread to wake up and do the processing. This is how I understood it.

app/models/manageiq/providers/nuage/network_manager/event_parser.rb
gberginc: Problem is this: https://github.com/manageiq/manageiq/blob/master/app/models/ems_event.rb#L243-L249. I incorrectly assumed this is because of the DB key, but here the filter will ignore the secon occurrence of the same event with the same timestamp.

bronaghs: @gberginc - Can you share an example of two different events that are being treated as duplicates?

miha-plesko:
@gberginc please close this PR i n favor of https://github.com/ManageIQ/manageiq-providers-nuage/pull/58 and https://github.com/ManageIQ/manageiq-providers-nuage/pull/59

PR #54, opened 2018-01-03
[WIP] Create event monitor only when amqp endpoints are defined

gasper-vrhovsek:
Ignore creating event monitor when no amqp endpoints are defined. Also a
bug when adding enpoints was fixed, where a default empty string was
added to the amqp endpoint list.

/cc @gberginc @miha-plesko 
@miq-bot add_label wip

This is linked to: https://bugzilla.redhat.com/show_bug.cgi?id=1527209

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
@@ -22,6 +22,15 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Runner < ManageI
     stop_event_monitor
   end
 
+  def do_work_loop
+    super if amqp?
+  end
+
+  def start_event_monitor
+    tid = super if amqp?
+    return tid unless tid.nil?
+  end
+
   def stop_event_monitor
     @event_monitor_handle.try!(:stop)
   ensure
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
```diff
@@ -22,6 +23,15 @@ def monitor_events
...
   end
 
+  def do_work_loop
+    super if has_amqp
+  end
+
+  def start_event_monitor
+    tid = super if has_amqp
+    return tid unless tid.nil?
+  end
```
miha-plesko: From what I understand we override the `start_event_monitor` function here so that when `has_amqp` is `false`, nothing is started, yes? If this is true, then I think no other change in this file is required, right? I mean, 

- do we then need to modify L3?
- do we then need to add L17?
- do we then need to also override `do_work_loop` in L26?

I may be missing something, but perhaps we can simplify this PR without affecting the functionality.

gasper-vrhovsek: Yes, it's a WIP. I remember doing this before the holiday season, and there were still some log outputs if I only prevented starting the event monitor. I'll write some tests and do more testing in development env.

miha-plesko:
@gasper-vrhovsek the issue has been resolved with https://github.com/ManageIQ/manageiq-providers-nuage/pull/69, please close this one.

gasper-vrhovsek:
Closing pull request.

PR #58, opened 2018-01-10
Upgrade qpid_proton related stuff to support v0.19.0

miha-plesko:
With this commit we adjust whatever needed to support qpid_proton version 0.19.0 instead of 0.18.1. Even if only a "minor" number has increased, the API is broken hence this commit. Most notable differences in qpid_proton API are:

- username/password can no longer be provided as part of URL, but separately
- class Qpid::Proton::Handler::MessagingHandler has been replaced with Qpid::Proton::Handler::MessagingHandler
- some handler callbacks have been renamed (on_start -> on_container_start etc.)
- some other functions have been renamed and signature has been updated
- `:urls` option is no longer supported (connection to fallback endpoint had to be implemented on MIQ level)

The reason to switch to qpid_proton 0.19.0 is that we measured huge performance issues on 0.18.1 that seem to be fixed in 0.19.0. Big thanks to @tadeboro for realizing that 0.19.0 was out and that it works better!

Fixes https://github.com/ManageIQ/manageiq-providers-nuage/issues/47
Must be merged together with: https://github.com/ManageIQ/manageiq/pull/16793

BZ https://bugzilla.redhat.com/show_bug.cgi?id=1534440
@miq-bot assign @juliancheal 
@miq-bot add_label gaprindashvili/yes, events

### DIFF
--- a/app/models/manageiq/providers/nuage/manager_mixin.rb
+++ b/app/models/manageiq/providers/nuage/manager_mixin.rb
@@ -81,10 +81,13 @@ module ManageIQ::Providers::Nuage::ManagerMixin
 
   def event_monitor_options
     @event_monitor_options ||= begin
+      amqp_auth = authentications.detect { |a| a.authtype == 'amqp' }
       {
         :ems                       => self,
         :urls                      => amqp_urls,
         :sasl_allow_insecure_mechs => true, # Only plain (insecure) mechanism currently supported
+        :user                      => amqp_auth.userid,
+        :password                  => amqp_auth.password
       }
     end
   end
@@ -93,12 +96,8 @@ module ManageIQ::Providers::Nuage::ManagerMixin
 
   def amqp_urls
     amqp_endpoints = endpoints.select { |e| e.role == 'amqp' || e.role.start_with?('amqp_fallback') }
-    amqp_auth      = authentications.detect { |a| a.authtype == 'amqp' }
-
     amqp_endpoints.map do |e|
-      url = "#{e.hostname}:#{e.port}"
-      url = "#{ERB::Util.url_encode(amqp_auth.userid)}:#{ERB::Util.url_encode(amqp_auth.password)}@#{url}" if amqp_auth
-      url
+      "#{e.hostname}:#{e.port}"
     end
   end
 
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
@@ -1,42 +1,48 @@
-class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler < Qpid::Proton::Handler::MessagingHandler
+class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler < Qpid::Proton::MessagingHandler
   def initialize(options = {})
     require 'qpid_proton'
 
     super()
     @options = options
 
-    @topics = @options.delete(:topics)
-    @test_connection = @options.delete(:test_connection)
+    @topics                = @options.delete(:topics)
+    @test_connection       = @options.delete(:test_connection)
     @message_handler_block = @options.delete(:message_handler_block)
+    @url                   = @options.delete(:url)
+    @timeout               = @options.delete(:amqp_connect_timeout) || 5.seconds
   end
 
-  def on_start(event)
-    @conn = event.container.connect(@options)
+  def on_container_start(container)
+    Timeout.timeout(@timeout) { @conn = container.connect(@url, @options) }
     unless @test_connection
-      @topics.each { |topic| event.container.create_receiver(@conn, :source => "topic://#{topic}") }
+      @topics.each { |topic| @conn.open_receiver("topic://#{topic}") }
     end
+  rescue Timeout::Error
+    raise MiqException::MiqHostError, "Timeout connecting to AMQP endpoint #{@url}"
   end
 
-  def on_connection_opened(event)
+  def on_connection_open(connection)
     # In case connection test was requested, close the connection immediately.
-    event.container.stop if @test_connection
+    connection.container.stop if @test_connection
   end
 
-  def on_connection_error(_event)
+  def on_connection_error(_connection)
     raise MiqException::MiqInvalidCredentialsError, "Connection failed due to bad username or password"
   end
 
-  def on_transport_error(_event)
-    # Only raise error if single URL is used, as otherwise qpid will attempt
-    # to fallback to alternative URLs.
-    raise MiqException::MiqHostError, "Transport error" unless @options[:urls].length > 1
+  def on_transport_error(_transport)
+    raise MiqException::MiqHostError, "Transport error"
   end
 
-  def on_message(event)
-    @message_handler_block.call(JSON.parse(event.message.body)) if @message_handler_block
+  def on_message(_delivery, message)
+    @message_handler_block&.call(JSON.parse(message.body))
+  end
+
+  def on_transport_close(_transport)
+    raise MiqException::MiqHostError, "Transport closed unexpectedly"
   end
 
   def stop
-    @conn.close
+    @conn&.close
   end
 end
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
@@ -2,17 +2,18 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream
   include Vmdb::Logging
 
   def self.test_amqp_connection(options = {})
+    return false if options[:urls].blank?
     # Ensure we just test the connection. AMQP channel will be established and
     # started, however it will be immediately stopped.
     options[:test_connection] = true
-    begin
-      stream = new(options)
+
+    stream = new(options)
+    ok = stream.with_fallback_urls(options[:urls]) do
       stream.connection.run
-      true
-    rescue => e
-      _log.info("#{log_prefix} Failed connecting to ActiveMQ: #{e.message}")
-      raise
+      return true
     end
+    raise MiqException::MiqHostError, "Could not connect to any of the #{options[:urls].count} AMQP hostnames" unless ok
+    true
   end
 
   def self.log_prefix
@@ -28,18 +29,40 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream
   def start(&message_handler_block)
     _log.debug("#{self.class.log_prefix} Opening amqp connection using options #{@options}")
     @options[:message_handler_block] = message_handler_block if message_handler_block
-    connection.run
+    with_fallback_urls(@options[:urls]) do
+      connection.run
+    end
   end
 
   def stop
-    @handler.stop
+    @handler&.stop
   end
 
   def connection
     unless @connection
-      @handler = ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler.new(@options)
-      @connection = Qpid::Proton::Reactor::Container.new(@handler)
+      @handler = ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler.new(@options.clone)
+      @connection = Qpid::Proton::Container.new(@handler)
     end
     @connection
+  end
+
+  def with_fallback_urls(urls)
+    urls.each_with_index do |url, idx|
+      endpoint_str = "ActiveMQ endpoint #{idx + 1}/#{@options[:urls].count} (#{url})"
+      _log.info("#{self.class.log_prefix} Connecting to #{endpoint_str}")
+      begin
+        @options[:url] = url
+        yield
+      rescue MiqException::MiqHostError, Errno::ECONNREFUSED, SocketError => err
+        _log.info("#{self.class.log_prefix} #{endpoint_str} errored: #{err}")
+        stop
+        reset_connection
+      end
+    end
+    false
+  end
+
+  def reset_connection
+    @connection = nil
   end
 end
--- a/bin/qpid_install.sh
+++ b/bin/qpid_install.sh
@@ -7,7 +7,7 @@ sudo apt-get install -y libsasl2-2 libsasl2-dev
 
 # Get the latest Qpid Proton source
 cd $HOME/build
-git clone --branch 0.18.1 https://github.com/apache/qpid-proton.git
+git clone --branch 0.19.0 https://github.com/apache/qpid-proton.git
 cd qpid-proton
 
 # Configure the source of Qpid Proton.
--- a/spec/models/manageiq/providers/nuage/network_manager_spec.rb
+++ b/spec/models/manageiq/providers/nuage/network_manager_spec.rb
@@ -54,7 +54,7 @@ describe ManageIQ::Providers::Nuage::NetworkManager do
     context 'AMQP connection' do
       before do
         @conn = double
-        allow(Qpid::Proton::Reactor::Container).to receive(:new).and_return(@conn)
+        allow(Qpid::Proton::Container).to receive(:new).and_return(@conn)
 
         creds = {}
         creds[:amqp] = {:userid => "amqp_user", :password => "amqp_password"}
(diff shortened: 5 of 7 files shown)


app/models/manageiq/providers/nuage/manager_mixin.rb
```diff
@@ -81,10 +81,13 @@ def verify_credentials(auth_type = nil, options = {})
 
   def event_monitor_options
     @event_monitor_options ||= begin
+      amqp_auth = authentications.detect { |a| a.authtype == 'amqp' }
```
gberginc: Is there a way to get into this without `amqp` auth type?

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -28,18 +33,39 @@ def initialize(options = {})
   def start(&message_handler_block)
     _log.debug("#{self.class.log_prefix} Opening amqp connection using options #{@options}")
     @options[:message_handler_block] = message_handler_block if message_handler_block
-    connection.run
+    with_fallback_urls(@options[:urls]) do
+      connection.run
```
gberginc: Does this effectively run multiple connections at the same time? Previously, fallback was handled by the Qpid itself.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -28,18 +33,39 @@ def initialize(options = {})
...
+    urls.each_with_index do |url, idx|
+      endpoint_str = "ActiveMQ endpoint #{idx + 1}/#{@options[:urls].count} (#{url})"
+      _log.info("#{self.class.log_prefix} Connecting to #{endpoint_str}")
+      begin
+        @options[:url] = url
+        yield
+      rescue MiqException::MiqHostError, Errno::ECONNREFUSED => err
+        _log.info("#{self.class.log_prefix} #{endpoint_str} errored: #{err}")
+        stop
+        reset_connection
```
gberginc: What happens when the first connection fails?

gberginc:
@miha-plesko double checking: the problem of multiple messages sent at the same time discussed [here](https://github.com/ManageIQ/manageiq-providers-nuage/pull/52/files#r157740110) still persists, right?

app/models/manageiq/providers/nuage/manager_mixin.rb
miha-plesko: We assume that if amqp credentials are not given, the worker won't be started. This is not the case yet, currently amqp empty credentials are inserted even when you pick "none" on the GUI in Events tab.

In other words: I'd leave it like this for now.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
miha-plesko: No, the `connection.run` line is blocking. Next fallback endpoint is only run when `connection.run` raises error. The only reason I've implemented auxilary `with_fallback_urls` is to improve readability, could have done it with a regular for loop as well :)

miha-plesko: You're correct, previously this was handled by ruby binding for qpid gem (thanks @tadeboro for digging into it). There is no such support in the actual qpid and since they changed the ruby binding to reflect C methods as much as possible, this must now be done on application level.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
miha-plesko: A very good question that I had to come to by hard :) It's working fine even when first connection fails, it proceeds to the fallback. I've tested with portforwarding as suggested by you some time ago. So I've set it to `localhost:5672, localhost:5672, <actual_endpoint>` and it worked even after EMS restart (that's why we rescue also `Errno::ECONNREFUSED`).

miha-plesko:
@gberginc see https://github.com/ManageIQ/manageiq-providers-nuage/pull/59. The issue there does not impact performance, it's just about making sure that no message is dropped if arriving at the same time as others.

gberginc:
@miha-plesko Please update `qpid_instal.sh` script as well. This is needed to pass the tests once the version bump in main repo is accepted.

https://github.com/ManageIQ/manageiq-providers-nuage/blob/master/bin/qpid_install.sh#L10

tadeboro:
Since I was the one doing the dumpster-diving for the CPU starvation problem, I will just give a quick recap of what happened between 0.18.1 and 0.19.0 qpid-proton releases that probably resolved the issue.

qpid-proton gained new proactor-based IO routers some time ago that batches the events received/sent. Running workload simulations on sample programs written in C failed to reproduce the problems (I tested qpid-proton versions 0.18.0, and 0.19.0), so I assumed that bindings are at fault here. A few `git grep`s and `git log`s later it became clear that all the good stuff did not make it to the bindings until the 0.19.0 release. And here we are now;)

gberginc:
@tadeboro thanks for looking into this issue. It’s really great knowing there is a reasonable solution in the end.

@alanconway can you confirm what @tadeboro is describing is correct?

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
gberginc: Two more:

* What happens when all three fail?
* What happens when one of the URLs will reappear?

gberginc:
> @gberginc see #59. The issue there does not impact performance, it's just about making sure that no message is dropped if arriving at the same time as others.

Sorry about missing the #59. Thanks for pointing it out.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
tadeboro: I cannot give a definite answer to what happens if all URLs are invalid, since my knowledge of this part of the code is really shallow, but my guess would be that the thread terminates and tells no one about it?

If we would like to achieve the same behavior as the old Ruby binding offered, then we should wrap lines 54:63 in infinite loop (old bindings created circular list from the urls that were passed to the connect call). Adding a loop will allow us to detect if/when one or more URLs reappear.

gberginc: I would kindly ask you to test this as it's an important feature that needs to either be implemented (loops with some delays, of course) or documented in the end.

miha-plesko: I've double tested and behavior is as expected: when certain endpoint dies it tries to connect to next fallback. If connection succeeds, it usines that fallback. If the fallback dies at some point, next callback is tested. If we're out of callbacks, then this is reported to log (same as when any previous endpoint died) and EventCatcher worker is restarted meaning that first endpoint is tested again.

Here is the console output that I get if I forcibly prevent Nuage from being able to access all three endpoints:
```
[----] ...::Nuage::NetworkManager::EventCatcher::Runner#start_event_monitor) EMS [...] as [...] Starting Event Monitor Thread
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) Connecting to ActiveMQ endpoint 1/3 (localhost:5672)
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) ActiveMQ endpoint 1/3 (localhost:5672) errored: Connection refused - connect(2) for "localhost" port 5672
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) Connecting to ActiveMQ endpoint 2/3 (localhost:5672)
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) ActiveMQ endpoint 2/3 (localhost:5672) errored: Connection refused - connect(2) for "localhost" port 5672
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) Connecting to ActiveMQ endpoint 3/3 (localhost:5672)
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) ActiveMQ endpoint 3/3 (localhost:5672) errored: Connection refused - connect(2) for "localhost" port 5672
[----] ...::Nuage::NetworkManager::EventCatcher::Runner#start_event_monitor) EMS [...] as [...] Started Event Monitor Thread
[----] ...::Nuage::NetworkManager::EventCatcher::Runner#do_work) EMS [...] as [...] Event Monitor Thread gone. Restarting...

[----] ...::Nuage::NetworkManager::EventCatcher::Runner#start_event_monitor) EMS [vsd01.phd.plm.nuagedemo.net] as [xlab] Starting Event Monitor Thread
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) Connecting to ActiveMQ endpoint 1/3 (localhost:5672)
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) ActiveMQ endpoint 1/3 (localhost:5672) errored: Connection refused - connect(2) for "localhost" port 5672
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) Connecting to ActiveMQ endpoint 2/3 (localhost:5672)
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) ActiveMQ endpoint 2/3 (localhost:5672) errored: Connection refused - connect(2) for "localhost" port 5672
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) Connecting to ActiveMQ endpoint 3/3 (localhost:5672)
[----] ...::Nuage::NetworkManager::EventCatcher::Stream#with_fallback_urls) MIQ(Class) ActiveMQ endpoint 3/3 (localhost:5672) errored: Connection refused - connect(2) for "localhost" port 5672
``` 

Then when I open tunnel again, effectively allowing EventCatcher to access ActiveMQ again, it connects immediately.

gberginc: This is perfect for me! Thanks @miha-plesko.

miha-plesko:
I've repushed this commit with two minor enhancements (compared to what was pushed earlier):

1) add `:amqp_connect_timeout` setting + the logic to apply it
2) forcibly restart event monitor on provider update

The former is needed because the new qpid_proton version has hard-coded timeout of _60 seconds_ which is too much. Especially when user clicks on "Validate" when creating/updating provider it's really annoying to have her wait `3 AMQP endpoints * 60 secs =180 seconds`!!! before reporting unaccessible endpoint.

The latter is needed because currently when user updated AMQP credentials the EventCatcher didn't get restarted. In other words, after providing new endpoints user had to restart the whole server to get it applied, which was horrible.

@gberginc I kindly ask for another pass when you can.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -2,17 +2,18 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream
   include Vmdb::Logging
 
   def self.test_amqp_connection(options = {})
+    return false if options[:urls].empty?
```
gberginc: If making such test, you should also check that `:urls` actually exists in `options`?

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -2,17 +2,18 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream
...
+    stream = new(options)
+    ok = stream.with_fallback_urls(options[:urls]) do
       stream.connection.run
-      true
-    rescue => e
-      _log.info("#{log_prefix} Failed connecting to ActiveMQ: #{e.message}")
-      raise
+      return true
     end
+    return ok if ok
```
gberginc: This looks interesting 😄. I'd make it either a bit more explicit, i.e. `return true if ok` or change the order, i.e.

```
raise MiqException::MiqHostError, "Could not ..." unless ok
return true
```

spec/models/manageiq/providers/nuage/network_manager_spec.rb
```diff
@@ -73,21 +73,6 @@
         expect { @ems.verify_credentials(:amqp) }.to raise_error(StandardError)
       end
     end
-
-    context 'credentials encoding' do
```
gberginc: We are removing this because it is no longer needed as creds are not part of AMQP URL, right?

app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
```diff
@@ -1,42 +1,48 @@
...
+    @test_connection       = @options.delete(:test_connection)
     @message_handler_block = @options.delete(:message_handler_block)
+    @url                   = @options.delete(:url)
+    @timeout               = @options.delete(:amqp_connect_timeout) || 5.seconds
   end
 
-  def on_start(event)
-    @conn = event.container.connect(@options)
+  def on_container_start(container)
+    Timeout.timeout(@timeout) { @conn = container.connect(@url, @options) }
```
gberginc: You are using the `Timeout` here because Qpid does not provide such mechanism on its own?

miha-plesko: Yes, the only timeout in whole qpid_proton ruby binding that I could find was:

```
 :idle_timeout (Numeric) — Seconds before closing an idle connection
```

which is not what we need here. So when I intentionaly input following AMQP urls when adding new provider:

```
google.com
yahoo.com
bbc.com
```

the UI got blocked for 3 minutes :D

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
miha-plesko: Done, thanks. Can't hurt to be onehundertpercent certain :)

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
miha-plesko: Wow, did I really code this? 😄  Fixed.

spec/models/manageiq/providers/nuage/network_manager_spec.rb
miha-plesko: Correct. Not only creds are not needed anymore as part of AMQP URL, but it's now even an error to include them there, so this test isn't valid anymore.

juliancheal:
Looking into this today :)

miha-plesko:
Thanks that's awesome @juliancheal because we're under uncomfortable time pressure 👍

bin/qpid_install.sh
```diff
@@ -7,7 +7,7 @@ sudo apt-get install -y libsasl2-2 libsasl2-dev
 
 # Get the latest Qpid Proton source
 cd $HOME/build
-git clone --branch 0.18.1 https://github.com/apache/qpid-proton.git
+git clone --branch 0.19.0 https://github.com/apache/qpid-proton.git
```
juliancheal: What's the reason why we're not using the Rubygem version for this? https://rubygems.org/gems/qpid_proton

miha-plesko: We are https://github.com/ManageIQ/manageiq/pull/16793/files. But unlike most other gems `qpid_proton` gem is special (yeeey) because it depends on system library `qpid-proton`.

```
$ gem install qpid_proton
```

The command above will fail unless `qpid-proton` is installed on your system. In other words, `qpid_proton` is just a Ruby binding for C program called `qpid-proton`. To get things even better, `qpid-proton` program is poorly supported by install tools (apt-get, yum) because precompiled packages are not prepared yet in official repos (they are somehow stuck at stonage versions).

juliancheal: Thanks @miha-plesko I knew there was a reason :)

miha-plesko:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1534440

juliancheal:
https://github.com/ManageIQ/manageiq/pull/16793 looks good to merge. I assume that one can be merged before this one?

gberginc:
Yes, version bump in the core must be merged before this one. The tests will fail otherwise as 0.19 introduced some API changes

juliancheal:
@gberginc Cool, I'll merge this as soon as core is merged.

gberginc:
Oh, that's just sooooo great!

Thanks @simaishi!
Thanks @juliancheal!
Thanks @miha-plesko!
And of course THANKS @tadeboro!

🍻 to all!

simaishi:

```
$ git log -1
commit a6a048df982c4a9c47ea10c870c0702655cf10b9
Date:   Mon Jan 22 19:59:14 2018 +0000

    Merge pull request #58 from miha-plesko/upgrade-to-qpid-0.19.0
    
    Upgrade qpid_proton related stuff to support v0.19.0
    (cherry picked from commit 51af0cbcf65e97b9f86099d63c9f2289278e2661)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1537301
```

alanconway:
Just back from vacation. This thread is long and it looks like some things
were figured out and fixed, is there still something I should do to help?

miha-plesko:
@alanconway AFAIK things are good here by now, thanks for asking.

PR #59, opened 2018-01-10
Assign :ems_ref to the event

miha-plesko:
With this commit we make sure that events are always created with unique ems_ref. For most of them we just grab value from the event hash, but we make up a new unique value when ID is not present in event hash to make is unique at least from this point on.

@miq-bot assign @bronaghs 

This is a replacement for https://github.com/ManageIQ/manageiq-providers-nuage/pull/52. There were two issues addressed in #52: the performance issue and the ID issue. The performance issue was solved by upgrading the underlying qpid_proton gem while the ID issue is solved by this PR.

/cc @gberginc

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/event_parser.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_parser.rb
@@ -8,7 +8,8 @@ module ManageIQ::Providers::Nuage::NetworkManager::EventParser
       :message    => event_type,
       :vm_ems_ref => nil,
       :full_data  => event.to_hash,
-      :ems_id     => ems_id
+      :ems_id     => ems_id,
+      :ems_ref    => event["requestID"].presence || "random-#{SecureRandom.uuid}"
     }
   end
 end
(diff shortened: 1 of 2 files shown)


gberginc:
@miq-bot add_label gaprindashvili/yes

bronaghs:
Looks good, I like this as a solution to the problem of false duplicate events. Can a spec test be written to cover this?

miha-plesko:
Thanks for pointing out @bronaghs. Implementing specs I've realized that it would be best to prefix those `ems_ref`s that were picked randomly by `random-` prefix. So that now ems_refs that we made up randomly now look like this:

```ruby
"random-#{RANDOM-UUID}"
```

e.g.

```
"random-56e69f6e-a1fd-457f-b4c5-844cfd790153"
```

This is just in order to be able to distinguish them later in case we need. @gberginc are you OK with this?

gberginc:
I am fine @miha-plesko.

simaishi:

```
$ git log -1
commit 87a7333fb420fb7cddd99a51143e4893b496b9a8
Author: Bronagh Sorota <bsorota@redhat.com>
Date:   Fri Jan 12 10:44:13 2018 -0500

    Merge pull request #59 from miha-plesko/add-event-id
    
    Assign :ems_ref to the event
    (cherry picked from commit 51c638ca53706fdee53694958d007f77cbfe226d)
```

miha-plesko:
Thanks for merging and backporting, I've opened a BZ for this (not sure if it was needed): https://bugzilla.redhat.com/show_bug.cgi?id=1534466

PR #60, opened 2018-01-12
Restart EventCatcher everytime we update provider

miha-plesko:
Prior this commit EventCatcher wasn't restarted when AMQP endpoints or fallback endpoints or port were updated, which is annoying. User had to either manually `sudo kill -9 PID` the old EventCatcher or restart the whole server.

With this commit we make sure that stop_event_monitor is called everytime when Nuage provider is updated.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1534460
@miq-bot assign @juliancheal 

/cc @gberginc

### DIFF
--- a/app/models/manageiq/providers/nuage/manager_mixin.rb
+++ b/app/models/manageiq/providers/nuage/manager_mixin.rb
@@ -79,6 +79,10 @@ module ManageIQ::Providers::Nuage::ManagerMixin
     end
   end
 
+  def credentials_changed?
+    true
+  end
+
   def event_monitor_options
     @event_monitor_options ||= begin
       {


gberginc:
@miha-plesko Can you please point to the code that you have shown to me (where this is called actually).

I honestly still have a problem with this as saying that credentials have changed whenever being asked, is something I would not expect.

@bronaghs do you perhaps have another idea how to overcome this?

miha-plesko:
https://github.com/miha-plesko/manageiq/blob/master/app/models/ext_management_system.rb#L636-L648

Piece of code above shows usage of `credentials_changed?` function usage that we're targeting with this PR. HOWEVER I must agree with @gberginc that this is not the best way to do it. Probably we should modify core repo instead to not only check whether default endpoint has changed, but also if amqp andpoint has (or any of its fallbacks).

miha-plesko:
We now have BZ for this PR: https://bugzilla.redhat.com/show_bug.cgi?id=1534460

bronaghs:
@miha-plesko Can you override [this](https://github.com/ManageIQ/manageiq/blob/master/app/models/ext_management_system.rb##L643..#L648) method in the Nuage Network Manager and check if the various AMQP endpoints,  fallback endpoints or ports have changed?

sasoc:
@miha-plesko Can you please close this PR, since I've made a [new one](https://github.com/ManageIQ/manageiq-providers-nuage/pull/65) that overrides `stop_event_monitor_queue_on_change` and fixes this issue.

miha-plesko:
Thanks @sasoc , closing.

PR #72, opened 2018-03-13
[WIP] Prevent AMQP connection from being closed periodically

miha-plesko:
With this commit we address two serious bugs that exist in qpid_proton gem:

1) Abruptly closed connections are leaking file descriptors
2) Connection gets abruptly closed every 2 minutes due to missing heartbeats

---

When AMQP connection is closed abruptly it results in TCP socket remaining open in CLOSE_WAIT state, which means file descriptor is not released:

```
$ ps -ef | grep MIQ
miha     108492     4234  3 12:02 pts/6    00:01:40 MIQ Server
miha   **108533** 108492  0 12:02 pts/6    00:00:07 MIQ: Nuage::NetworkManager::EventCatcher id: 105, queue: ems_3
miha     108545   108492  0 12:02 pts/6    00:00:02 MIQ: MiqEventHandler id: 106, queue: ems
miha     108554   108492  0 12:02 pts/6    00:00:04 MIQ: MiqGenericWorker id: 107, queue: generic

$ lsof -ap 108533 | grep CLOSE_WAIT
ruby    108533 miha  116u  IPv4   562438  0t0  TCP 172.16.117.189:53626->147.75.102.132:amqp (CLOSE_WAIT)
ruby    108533 miha  197u  IPv4   561644  0t0  TCP 172.16.117.189:53630->147.75.102.132:amqp (CLOSE_WAIT)
ruby    108533 miha  311u  IPv4   560657  0t0  TCP 172.16.117.189:53634->147.75.102.132:amqp (CLOSE_WAIT)
ruby    108533 miha  549u  IPv4   565342  0t0  TCP 172.16.117.189:53642->147.75.102.132:amqp (CLOSE_WAIT)
ruby    108533 miha  576u  IPv4   565122  0t0  TCP 172.16.117.189:53650->147.75.102.132:amqp (CLOSE_WAIT)
... (keeps growing)
```

After period of time (9 hours in our case) there are enough file descriptors open for operating system to yield:

```
Too many open files - socket(2) for "172.16.117.189" port 5672
```

This is a bug in qpid_proton gem as reported here: https://issues.apache.org/jira/browse/PROTON-1791 Until it gets resolved, we're introducing a workaround for it with this commit. Basically we capture socket references and manually close them upon closing AMQP connection.

With this commit we also fix a problem in qpid_proton gem that is not sending heartbeat to the  ActiveMQ. Then ActiveMQ thinks we're dead and closes connection abruptly. Solution for this problem is as simple as poking AMQP connection every 5 seconds so that ActiveMQ know we're still alive.

@miq-bot assign @juliancheal 

/cc @gberginc

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
@@ -13,6 +13,7 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler
   end
 
   def on_container_start(container)
+    $nuage_log.debug("#{self.class.log_prefix} Starting AMQP")
     Timeout.timeout(@timeout) { @conn = container.connect(@url, @options) }
     unless @test_connection
       @topics.each { |topic| @conn.open_receiver("topic://#{topic}") }
@@ -26,8 +27,10 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler
     connection.container.stop if @test_connection
   end
 
-  def on_connection_error(_connection)
-    raise MiqException::MiqInvalidCredentialsError, "Connection failed due to bad username or password"
+  def on_connection_error(connection)
+    msg = "#{self.class.log_prefix} AMQP on_connection_error: #{connection.condition}"
+    $nuage_log.debug(msg)
+    raise MiqException::MiqHostError, msg
   end
 
   def on_transport_error(_transport)
@@ -43,6 +46,14 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler
   end
 
   def stop
-    @conn&.close
+    unless @conn.nil? || @conn.container.stopped
+      $nuage_log.debug("#{self.class.log_prefix} Stopping AMQP")
+      @conn.container.stop
+    end
+    @conn = nil
+  end
+
+  def self.log_prefix
+    "MIQ(#{name})"
   end
 end
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/qpid_container.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/qpid_container.rb
@@ -1,0 +1,80 @@
+# NOTE: here we patch Qpid::Proton::Container object to tackle two blocking bugs that we observed:
+# a) sockets are not being closed properly upon connection close which resulty in file descriptor leakage
+# b) AMQP heartbeat is not being sent which results in periodically closing connection every minute or two
+# See https://issues.apache.org/jira/browse/PROTON-1791
+# TODO: remove this entire class once the issues are fixed (should be in qpid_proton 0.22.0)
+
+class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::QpidContainer < Qpid::Proton::Container
+  # Override to capture connection drivers.
+  def connection_driver(io, opts = nil, server = false)
+    driver = super(io, opts, server)
+
+    # Capture connection drivers so that we can manually poke them later.
+    @drivers ||= []
+    @drivers << driver
+
+    driver
+  end
+
+  # Override to also start poking thread upon container start.
+  def run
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/nuage/network_manager/event_catcher/qpid_container.rb
```diff
@@ -0,0 +1,15 @@
...
+  def connection_driver(io, opts = nil, server = false)
+    # Store references to sockets so that we can later close them.
+    @sockets ||= []
+    @sockets << io
+
+    super(io, opts, server)
+  end
+
+  def close_sockets
+    (@sockets || []).each(&:close)
```
gberginc: Should you also clear the array after sockets are closed? Is the container reused later?

app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
```diff
@@ -21,13 +22,20 @@ def on_container_start(container)
...
   def on_connection_open(connection)
     # In case connection test was requested, close the connection immediately.
     connection.container.stop if @test_connection
   end
 
-  def on_connection_error(_connection)
-    raise MiqException::MiqInvalidCredentialsError, "Connection failed due to bad username or password"
+  def on_connection_error(connection)
+    c = connection.condition
+    msg = "#{self.class.log_prefix} AMQP on_connection_error: #{c}"
```
gberginc: Can you condense l35 and l36 into a single one, since `c` is not used anywhere else.

app/models/manageiq/providers/nuage/network_manager/event_catcher/qpid_container.rb
miha-plesko: Good point. Container is generally not reused (a new instance is always created [here](https://github.com/ManageIQ/manageiq-providers-nuage/blob/master/app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb#L59)), but doing it properly is good idea anyway.

app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
miha-plesko: Done, thanks

miha-plesko:
@gberginc thanks for review. I've pushed a slightly modified code. Most notably, we're now manually closing sockets in container's `stop` method as opposed to doing it in `on_container_stop` before. It then turned out that `container.stop` method was never actually invoked! (we were only closing the connection `connection.close`, which I think is wrong, should be stopping container instead). Therefore I've updated the code to stop container, not just close connection and let container close all the sockets.

Unfortunately, even when invoking `container.stop`, stalled tcp sockets persisted, therefore I had to keep the code that closes them manually.

juliancheal:
@gberginc You happy with this PR?

miha-plesko:
I'm just about to push updated version, sorry for not marking it WIP. We had a conversation with qpid_proton authors and I'm preparing a slightly enhanced patch (just need to comment it and I push).

https://issues.apache.org/jira/browse/PROTON-1791

That being said, I'd be happy if we then get it merged today or tomorrow, because it's kind of a blocking bug.

juliancheal:
Cool. Let's have a look when you push the updates :)

miha-plesko:
@juliancheal @gberginc I've pushed the new patch. Testing it locally I can see that AMQP connection now remains established (for at least 15minutes that I've tested). Also, when intentionally closing it abruptly it didn't leak any file descriptors.

Please note that as soon as the two issues are fixed in qpid_proton gem we'll just remove entire `app/models/manageiq/providers/nuage/network_manager/event_catcher/qpid_container.rb` file to use the official solution.

I kindly ask you guys for a fast review; we're kind of a in a hurry to catch the next release...

miha-plesko:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-nuage/pull/76 where we only focus on fixing the file descriptor leak. Hopefully a new version of qpid_proton gem will be out in a few days with heartbeating problem properly resolved (see https://issues.apache.org/jira/browse/PROTON-1782).

PR #74, opened 2018-03-13
[WIP] Silently reconnect qpid in case of heartbeat timeout

miha-plesko:
We observed that Nuage server often "forgets" to send AMQP heartbeat for some reason which results in Nuage provider's event catcher reporting the error in evm.log and reconnecting to the next fallback url.

With this commit we capture such type of exception and reconnect silently to the same fallback url to prevent event catcher from looping fallback urls for this error and also to keep evm.log clear.

@miq-bot assign @juliancheal 

/cc @gberginc

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
@@ -4,6 +4,7 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Runner < ManageI
       options = @ems.event_monitor_options
       options[:topics] = worker_settings[:topics]
       options[:amqp_connect_timeout] = worker_settings[:amqp_connect_timeout]
+      options[:idle_timeout] = worker_settings[:amqp_idle_timeout]
       @event_monitor_handle = ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream.new(options)
     end
     @event_monitor_handle
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
@@ -17,7 +17,7 @@ class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::Stream
   end
 
   def self.log_prefix
-    "MIQ(#{self.class.name})"
+    "MIQ(#{self.name})"
   end
 
   def initialize(options = {})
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -30,7 +30,7 @@ def start(&message_handler_block)
     $nuage_log.debug("#{self.class.log_prefix} Opening amqp connection using options #{@options}")
     @options[:message_handler_block] = message_handler_block if message_handler_block
     with_fallback_urls(@options[:urls]) do
-      connection.run
+      run_with_idle_timeout_ignored
```
gberginc: I would not name the method like this. Isn't `run` ok and implement whatever logic there is needed? Imagine we will realise there is another error message that we ignore and you will have to rename it to `run_with_iddle_timeout_and_one_other_error_ignored` 😄.

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -62,6 +62,23 @@ def with_fallback_urls(urls)
...
+  # TODO: Don't ignore idle timeout error when following bug in qpid_proton is fixed:
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1418026
+  # Until then, we don't want this error to keep restarting our event catcher in case there are no events emitted
+  # for 16 seconds.
+  def run_with_idle_timeout_ignored
+    loop do
+      begin
+        connection.run
+      rescue MiqException::MiqHostError => e
+        raise unless e.message.include?('local-idle-timeout expired')
```
gberginc: I presume there is no "error code" or something similar that you could check for?

app/models/manageiq/providers/nuage/network_manager/event_catcher/runner.rb
```diff
@@ -4,6 +4,7 @@ def event_monitor_handle
       options = @ems.event_monitor_options
       options[:topics] = worker_settings[:topics]
       options[:amqp_connect_timeout] = worker_settings[:amqp_connect_timeout]
+      options[:idle_timeout] = worker_settings[:amqp_idle_timeout]
```
gberginc: Does this setting make sense at all?

app/models/manageiq/providers/nuage/network_manager/event_catcher/stream.rb
```diff
@@ -62,6 +62,23 @@ def with_fallback_urls(urls)
...
+        $nuage_log.debug("#{self.class.log_prefix} Silently reconnect on 'local-idle-timeout expired'")
+        stop
```
gberginc: This effectively closes the connection on the [messaging handler](https://github.com/miha-plesko/manageiq-providers-nuage/blob/8f74cab92dfdae1614ea1f6b186bd0e9cc6fab70/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb#L46). Does this not occur automatically in case of an exception?

juliancheal:
@gberginc are you happy with this PR to have the WIP flag removed?

miha-plesko:
Closing if favor of https://github.com/ManageIQ/manageiq-providers-nuage/pull/72. The patch there will address both issues (file descriptor leakage and the root cause for connection closing which resulted in fd leakage)

PR #77, opened 2018-03-16
Move qpid_proton dependency from core repo to nuage repo

miha-plesko:
With this commit we introduce qpid_proton dependency directly here on Nuage provider. It was originally put to the core repo because it was only installable from github repository. This is no longer the case as gem is available in RubyGems repository:

https://rubygems.org/gems/qpid_proton

We want to be able to shift gem version when Nuage provider integration proceeds without affecting the core repo therefore it's better for it to reside here.

(BZ above is the reason why we want to shift qpid proton version to 0.22.0, but in the intermediate step the core version must match with plugin's version or else we get `Bundler could not find compatible versions`)

@miq-bot assign @juliancheal

/cc @gberginc

### DIFF
--- a/manageiq-providers-nuage.gemspec
+++ b/manageiq-providers-nuage.gemspec
@@ -14,6 +14,7 @@ Gem::Specification.new do |s|
   s.files = Dir["{app,config,lib}/**/*"]
 
   s.add_runtime_dependency "excon",       "~>0.40"
+  s.add_runtime_dependency "qpid_proton", "~>0.19.0"
 
   s.add_development_dependency "codeclimate-test-reporter", "~> 1.0.0"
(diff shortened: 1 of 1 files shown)


gberginc:
LGTM

juliancheal:
@Fryguy would this PR be ok? Or does it cause the same issue as https://github.com/ManageIQ/manageiq/pull/17163#issuecomment-373760492

miha-plesko:
@juliancheal either both PRs must be merged or none. Based on comment from @Fryguy I must yield from this PR and agree that leaving `qpid_proton` gem optional is ATM still the best option after all (see [here](https://github.com/ManageIQ/manageiq/pull/17163#issuecomment-374507101)), therefore I'm closing.

PR #89, opened 2018-05-23
Cleanup config/settings.yaml after changes in UI

miha-plesko:
With this commit we cleanup settings.yaml since UI no longer needs `::Settings.product.nuage` setting but rather relies on config/permissions.yaml entries.

Depends on: https://github.com/ManageIQ/manageiq-ui-classic/pull/3979

/cc @djberg96

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -14,8 +14,6 @@
   :nuage_network:
     :inventory_object_refresh: true
     :allow_targeted_refresh: true
-:product:
-  :nuage: true
 :workers:
   :worker_base:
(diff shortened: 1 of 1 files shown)


miha-plesko:
Marking as WIP because it needs to be merged *after* the UI PR.
Marking as GA/no because we don't really need to backport this harmless cleanup, it's an investment for future :)

djberg96:
:+1:

miha-plesko:
Dependant PR just got merged, @juliancheal I think we're ready to go.

PR #90, opened 2018-05-25
Drop legacy refresh parser

miha-plesko:
With this commit we delete legacy refresher because it has not been used for a while now and would only cause problems now that we're about to extend the new parser with more models.

RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1574903


miha-plesko:
@juliancheal it think we're good to go on this one.

miha-plesko:
Rebased, assigning to @Ladas since you've reviewed it.

PR #92, opened 2018-05-28
Inventory Enterprise (Nuage) into CloudTenant (MIQ)

miha-plesko:
With this commit we inventory Enterprise into `CloudTenant` model instead into `NetworkGroup` model. We capture its name and description and hook it to the network manager.

Followup PR on UI: https://github.com/ManageIQ/manageiq-ui-classic/pull/3995

RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1574903

@Ladas with this commit we stop using `NetworkGroup` and use `CloudTenant` instead. We promised such change [long time ago](https://github.com/ManageIQ/manageiq/pull/16136#issuecomment-334778694) and here we are. This commit seems large but what it does in fact is a find-and-replace

```
"network_group" -> "cloud_tenant"
```

and nothing more. Most of the files are in fact specs...


Ladas:
@miha-plesko interesting, I think NetworkGroup was originally designed to replace CloudNetwork model. We did a new model, because it was a bit different. Does CloudTenant fit this better?

miha-plesko:
@Ladas I'm pretty sure, yes. See here how Enterprise object looks like: https://nuagenetworks.github.io/vsd-api-documentation/v5_0/enterprise.html - it's more like a "project" i.e. "tenant" i.e. "scope" than a network entity, because it has no networking properties like CIDR.

We had a discussion with @gtanzillo , @Fryguy and @gberginc about this and I think CloudTenant seems the most reasonable choice here. Looking forward to your opinion though.

app/models/manageiq/providers/nuage/inventory/collector/network_manager.rb
```diff
@@ -28,8 +28,8 @@ def security_group(ems_ref)
     security_groups.find { |sg| sg['ID'] == ems_ref }
   end
 
-  def network_group(ems_ref)
-    network_groups.find { |ng| ng['ID'] == ems_ref }
+  def cloud_tenant(ems_ref)
+    cloud_tenants.find { |tenant| tenant['ID'] == ems_ref }
```
Ladas: we could cache it like in target_collection.rb, so we don't do repeated O(n) searches

miha-plesko:
Thanks for suggesting the caching improvement, I've updated the code. I can prepare a PR in core to remove NetworkGroup completely @Ladas if you're sure no other provider uses it.

Ladas:
@miha-plesko I am 98% sure no other provider uses NetworkGroup :-) We've created it just for Nuage and the work for using it further was never finished.

miha-plesko:
@Ladas OK I'll remove it then. Are you good with this PR now? We want to further expand inventory and need this one merged to continue.

PR #93, opened 2018-05-29
Inventory L3 Domain (Nuage) into NetworkRouter (MIQ)

miha-plesko:
With this commit we inventory Nuage L3 Domain entity into NetworkRouter model. Existing refresher already depended on some data from L3 Domain entities but didn't inventory them properly. Dirty-hack with extra_attributes hash on CloudSubnet was employed... well, not anymore - with this commit we do it properly.

Nuage NetworkManager summary page:
capture

Selected NetworkRouter summary page:
capture

RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1574920


miha-plesko:
@Ladas I think this is the last "ffs not 15 files again" kind of PR for Nuage, because Domains were incorrectly inventoried. When we merge this, we'll start introducing new entities (FloatingIP, NetworkPort) so PRs will be smaller... Bear with us, please 🍬

miha-plesko:
I think the codeclimate errors can be ignored since we're following pattern.

miha-plesko:
@Ladas pingo pongo

PR #94, opened 2018-05-31
Inventory L2 Domain entity (Nuage) as CloudSubnet model (MIQ)

miha-plesko:
With this commit we introduce a subclass of CloudSubnet model that represents a standalone subnet, called "L2 domain" on Nuage. Main difference is that such standalone domain is not connected to any NetworkRouter.

We're introducing `CloudSubnetL2` because it's important for Nuage that they are able to differentiate between the two (L3 subnet and L2 subnet). Hence we also introduce auxilary methods `l3_cloud_subnets` and `l2_cloud_subnets` on NetworkManager and CloudTenant model to be able to only get desired subnet type.

RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1574921

### DIFF
--- a/app/models/manageiq/providers/nuage/inventory/collector.rb
+++ b/app/models/manageiq/providers/nuage/inventory/collector.rb
@@ -9,11 +9,12 @@ class ManageIQ::Providers::Nuage::Inventory::Collector < ManagerRefresh::Invento
   end
 
   def initialize_inventory_sources
-    @cloud_tenants   = {}
-    @cloud_subnets   = []
-    @security_groups = []
-    @zones           = {}
-    @network_routers = {}
+    @cloud_tenants    = {}
+    @cloud_subnets    = []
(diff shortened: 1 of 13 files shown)


miha-plesko:
@Ladas so in this PR we inventory subnets like this:

```
L3 subnet ---> ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet
L2 subnet ---> ManageIQ::Providers::Nuage::NetworkManager::CloudSubnetL2
```

I like such approach because it's easy to filter out the subnet type we want but I'm curious whether you'd find it too hacky?

miha-plesko:
@juliancheal or @Ladas looking forward to your review.

juliancheal:
LGTM 👍 @Ladas any comments from you.

miha-plesko:
Migration is provided here https://github.com/ManageIQ/manageiq-schema/pull/215

PR #97, opened 2018-06-07
Inventory FloatingIps

miha-plesko:
With this commit we inventory FloatingIPs. There's a catch: we want them to be connected to NetworkRouter directly, not via CloudNetwork, so we need to add additional field network_router_id for this (see related PR) and override how Nuage's NetworkRouter queries for its FloatingIps (to not go through CloudNetworks).

RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1574922

Depends on: https://github.com/ManageIQ/manageiq-schema/pull/217


app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
```diff
@@ -37,6 +37,10 @@ def network_routers
     @network_routers_map.values.compact
   end
 
+  def floating_ips
+    [] # TODO(miha-plesko): implement targeted refresh for floating ips
```
gberginc: Please create issues for this so that we don't forget on them. I see another TODO above.

app/models/manageiq/providers/nuage/inventory/parser/network_manager.rb
```diff
@@ -65,6 +66,16 @@ def security_groups
     end
   end
 
+  def floating_ips
+    collector.floating_ips.each do |ip|
+      persister.floating_ips.find_or_build(ip['ID']).assign_attributes(
+        :address        => ip['address'],
+        :network_router => persister.network_routers.lazy_find(ip['parentID']),
```
gberginc: Can you explain this bit? Why both `network_router` and `cloud_tenant` use the same data from the floating IP?

app/models/manageiq/providers/nuage/network_manager/vsd_client/rest.rb
```diff
@@ -32,6 +32,13 @@ def append_headers(key, value)
     @headers[key] = value
   end
 
+  def clear_headers
```
gberginc: May I suggest to use a different method name? `init_headers`, `reset_headers` or something similar. I had a comment asking why are you clearing the headers [here](https://github.com/ManageIQ/manageiq-providers-nuage/pull/97/files#diff-f361099d477f41489f9fa7843a091b63R100). After I saw this, I understand that you are basically resetting them.

miha-plesko: Good point, thanks. In fact I decided to open a separate PR to fix the vsd client itself so that this one can focus on FloatingIp inventoring only.

https://github.com/ManageIQ/manageiq-providers-nuage/pull/99

app/models/manageiq/providers/nuage/inventory/parser/network_manager.rb
gberginc: Oh, I missed the fact that you query the network router for its `cloud_tenant` in the next line. Thanks!

miha-plesko:
@Ladas I've updated the code to leverage CloudNetwork when relating FloatingIp to NetworkRouter. HOWEVER there are two issues with such approach:

a) on Nuage GUI you can allocate floating IP from external network **to a router (which can then allocate it to vPort beneath it)**. With our model here this is not possible: FloatingIp is connected to external network but MIQ is not aware of what router is it allocated to. So I think we still need to add that additional field to FloatingIp model.

b) on Nuage GUI you can allocate floating IPs from **different external networks** to the same router. With our model, this is not possible, you can only assign a single one.

We can have a call tomorrow so that I show you how it looks like on Nuage GUI. Would be great to push this through this week so that we can then model also NetworkPorts and be done with inventory upgrade.

Ladas:
@miha-plesko 
a) ok, looks like we need network_router_id in FloatingIp

b) so in Nuage, do you do some explicit connection of a Router and External network? Sounds like the connection is done only by allocation the floating IP to router (meaning you can pick from any external network?). So might be enough that the floating_ip has info from what network and router it comes from.

miha-plesko:
Yes @Ladas you're correct about the b) - there is no explicit connection between the router and external netrwork, only FloatingIp itself is aware of what external network did it come from. I'm going to update this PR to match following behavior:

- inventory exernal networks as CloudNetwork
- inventory NetworkRouters (no connection to CloudNetwork)
- inventory FloatingIps and connect them to both NetworkRouter and CloudNetwork

Then I think we're good.

miha-plesko:
There you go @Ladas, I've updated the PR and I think relations between routers, floating ips and external natworks are now in great shape. Looking forward to your review. Will need to wait shema update before we merge.

miha-plesko:
@Ladas I've commented out the line where we bind FloatingIp to NetworkRouter and put TODO there so that we can, should you agree, merge this PR today. Once the schema PR will get merged I'll uncomment them. Would this work for you?

Ladas:
@miha-plesko sounds good :+1:

PR #99, opened 2018-06-08
Reset API headers after each call

miha-plesko:
With this commit we modify VsdClient so that each API call is performed with default headers unless explicitly speficied differently.

Having the editor open I also updated the code in VsdClient::Rest because it contained a copy-paste of entire sections instead reusing central function...

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/vsd_client.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/vsd_client.rb
@@ -93,8 +93,11 @@ module ManageIQ::Providers
 
     def get_list(url)
       response = @rest_call.get("#{@server}/#{url}")
-      return unless response.code == 200
-      return [] if response.body.empty?
+
+      # Reset headers after call was performed or next API call will send them too.
+      @rest_call.reset_headers
+
+      return [] if response.body.empty? || response.code != 200
       JSON.parse(response.body)
     end
 
--- a/app/models/manageiq/providers/nuage/network_manager/vsd_client/rest.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/vsd_client/rest.rb
@@ -5,7 +5,7 @@ class ManageIQ::Providers::Nuage::NetworkManager::VsdClient::Rest
     @user = user
     @password = password
     @api_key = ''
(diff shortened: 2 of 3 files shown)


miha-plesko:
@gberginc green

app/models/manageiq/providers/nuage/network_manager/vsd_client.rb
```diff
@@ -93,6 +93,13 @@ def exclude_name(name)
 
     def get_list(url)
       response = @rest_call.get("#{@server}/#{url}")
+
+      # Reset headers after call was performed or next API call will send them too.
+      @rest_call.reset_headers
+
+      # API returns 204 No Content if too strict filter is defined.
+      return [] if response.code == 204
```
gberginc: May I suggest you merge this with L104, i.e.

```
return [] if response.body.empty? or response.code == 204
```

app/models/manageiq/providers/nuage/network_manager/vsd_client/rest.rb
```diff
@@ -32,46 +32,40 @@ def append_headers(key, value)
...
-      login
-    end
+    request(url, :method => :post, :data => data)
+  end
 
-    RestClient::Request.execute(:method => :post, :data => data, :url => url, :user => @user, :password => @api_key,
-    :headers => @headers, :verify_ssl => false) do |response|
-      return response
-    end
+  def request(url, method: :get, data: nil, verify: false)
```
gberginc: This is really a nice catch.

app/models/manageiq/providers/nuage/network_manager/vsd_client.rb
miha-plesko: Good point. I actually modified behavior to always return empty list `[]` unless status is 200 and body is present. Not sure why we complicated in first place.

app/models/manageiq/providers/nuage/network_manager/vsd_client/rest.rb
gberginc: I suggest you use `verify_ssl` as the name of the argument. Just `verify` is a bit misleading to me.

miha-plesko: Done

PR #101, opened 2018-07-12
Inventory vPort entities (Nuage) into NetworkPort model (MIQ)

miha-plesko:
With this commit we inventory NetworkPorts that can be of different
types in Nuage:

- Bridge vPort
- Host vPort
- VM vPort
- Container vPort

Each of them has a unique way of retrieving IP address by performing additional API request, hence the case-when statement in the parser.

image

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1574927

### DIFF
--- a/app/models/manageiq/providers/nuage/inventory/persister/definitions/network_collections.rb
+++ b/app/models/manageiq/providers/nuage/inventory/persister/definitions/network_collections.rb
@@ -7,11 +7,20 @@ module ManageIQ::Providers::Nuage::Inventory::Persister::Definitions::NetworkCol
        security_groups
        cloud_networks
        floating_ips
+       network_ports
        network_routers).each do |name|
 
       add_collection(network, name) do |builder|
         builder.add_properties(:parent => manager) # including targeted
       end
+    end
+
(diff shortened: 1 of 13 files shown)


app/models/manageiq/providers/nuage/inventory/persister/definitions/network_collections.rb
```diff
@@ -7,11 +7,20 @@ def initialize_network_inventory_collections
...
         builder.add_properties(:parent => manager) # including targeted
       end
     end
+
+    add_cloud_subnet_network_ports
+  end
+
+  def add_cloud_subnet_network_ports(extra_properties = {})
+    add_collection(network, :cloud_subnet_network_ports, extra_properties) do |builder|
+      builder.add_properties(:parent => manager, :parent_inventory_collections  => %i(network_ports))
```
Ladas: could you fix this 1 rubocop nit? There is extra space

miha-plesko:
Yeah, had to run somewhere yesterday then I didn't come back until today. Sorry it took me a day to remove this single space.

PR #102, opened 2018-07-13
Temporarily remove many-to-many from NetworkRouter

miha-plesko:
With this commit we remove relation which allows NetworkRouter to relate to multiple FloatingIps directly (instead via CloudNetwork). Until https://github.com/ManageIQ/manageiq-schema/pull/217 is merged we are not able to use such direct relation or else UI crashes.

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/network_router.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/network_router.rb
@@ -1,3 +1,8 @@
 class ManageIQ::Providers::Nuage::NetworkManager::NetworkRouter < ::NetworkRouter
   has_many :floating_ips, :dependent => :destroy
+
+  # TODO(miha-plesko): remove when https://github.com/ManageIQ/manageiq-schema/pull/217 is merged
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/nuage/network_manager/network_router.rb
```diff
@@ -1,3 +1,4 @@
 class ManageIQ::Providers::Nuage::NetworkManager::NetworkRouter < ::NetworkRouter
-  has_many :floating_ips, :dependent => :destroy
+  # TODO(miha-plesko): uncomment when https://github.com/ManageIQ/manageiq-schema/pull/217 is merged
+  # has_many :floating_ips, :dependent => :destroy
```
Ladas: the best might be to define:

```ruby
def floating_ips
  FloatingIp.none
end
```

so you can have it used in UI or other places, but it will just return blank result until properly implemented

miha-plesko: Makes sense, done.

PR #103, opened 2018-07-16
[DO NOT MERGE] Initial support for AnsibleRunner-driven CRUD

miha-plesko:
With this commit we provide initial support for AnsibleRunner-driven CRUD operations. We provide following ice-breaking examples:

- DELETE (cloud subnet, floating ip, security group)
- CREATE (cloud subnet)

The DELETE functionality is fully functional with this PR, but the CREATE functionality isn't yet, because it needs UI changes.

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager.rb
+++ b/app/models/manageiq/providers/nuage/network_manager.rb
@@ -91,4 +91,53 @@ class ManageIQ::Providers::Nuage::NetworkManager < ManageIQ::Providers::NetworkM
   def l2_cloud_subnets
     cloud_subnets.where(:type => self.class.l2_cloud_subnet_type)
   end
+
+  def ansible_env_vars
+    {}
+  end
+
+  def ansible_extra_vars(extra)
+    {
+      :nuage_auth => {
+        :api_username   => default_authentication.userid,
+        :api_password   => default_authentication.password,
+        :api_enterprise => 'csp',
+        :api_version    => api_version.sub('.', '_'),
+        :api_url        => self.class.base_url(
+          default_endpoint.security_protocol,
+          default_endpoint.hostname,
+          default_endpoint.port
+        )
+      }
+    }.merge(extra)
+  end
+
+  def ansible_root
+    ManageIQ::Providers::Nuage::Engine.root.join("content_tmp/ansible")
+  end
+
+  def playbook(name)
+    ansible_root.join(name)
+  end
+
+  def create_cloud_subnet_queue(userid, options = {})
+    task_opts = {
+      :action => "creating Cloud Subnet for user #{userid}",
+      :userid => userid
+    }
+    queue_opts = {
+      :class_name  => self.class.name,
+      :method_name => 'create_cloud_subnet',
+      :instance_id => id,
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => 'ems_operations',
+      :zone        => my_zone,
+      :args        => [options]
+    }
+    MiqTask.generic_action_with_callback(task_opts, queue_opts)
+  end
+
+  def create_cloud_subnet(options)
+    CloudSubnet.raw_create_cloud_subnet(self, options)
+  end
 end
--- a/app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
@@ -1,2 +1,60 @@
 class ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet < ::CloudSubnet
+  supports :delete
+  supports :create
+
+  def delete_cloud_subnet_queue(userid)
+    task_opts = {
+      :action => "deleting Cloud Subnet for user #{userid}",
+      :userid => userid
+    }
+    queue_opts = {
+      :class_name  => self.class.name,
+      :method_name => 'raw_delete_cloud_subnet',
+      :instance_id => id,
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => 'ems_operations',
+      :zone        => ext_management_system.my_zone,
+      :args        => []
+    }
+    MiqTask.generic_action_with_callback(task_opts, queue_opts)
+  end
+
+  def raw_delete_cloud_subnet
+    $nuage_log.info("Deleting Cloud Subnet (ems_ref = #{ems_ref})")
+
+    Ansible::Runner.run(
+      ext_management_system.ansible_env_vars,
+      ext_management_system.ansible_extra_vars(:id => ems_ref),
+      ext_management_system.playbook('remove-subnet.yml')
+    )
+
+    $nuage_log.info("Done deleting Cloud Subnet (ems_ref = #{ems_ref})")
+  rescue => e
+    $nuage_log.error "Error deleting Cloud Subnet: #{e}"
+    raise MiqException::MiqCloudSubnetDeleteError
+  end
+
+  def self.raw_create_cloud_subnet(ext_management_system, options)
+    $nuage_log.info("Create Cloud Subnet (ems_ref = #{ems_ref})")
+
+    # TODO: update UI to match what Nuage needs
+
+    subnet = Ansible::Runner.run(
+      ext_management_system.ansible_env_vars,
+      ext_management_system.ansible_extra_vars(
+        :domain_id => options[:router_id],
+        :subnet_attributes => { :name => options[:name] }
+      ),
+      ext_management_system.playbook('remove-subnet.yml')
+    )
+
+    $nuage_log.info("Done creating Cloud Subnet (ems_ref = #{ems_ref})")
+
+    # TODO: can playbook return value?
+
+    {:ems_ref => subnet.id, :name => options[:name]}
+  rescue => e
+    $nuage_log.error "Error creating Cloud Subnet: #{e}"
+    raise MiqException::MiqCloudSubnetCreateError
+  end
 end
--- a/app/models/manageiq/providers/nuage/manager_mixin.rb
+++ b/app/models/manageiq/providers/nuage/manager_mixin.rb
@@ -19,9 +19,13 @@ module ManageIQ::Providers::Nuage::ManagerMixin
       end
     end
 
-    def auth_url(protocol, server, port, version)
+    def base_url(protocol, server, port)
       scheme = %w(ssl ssl-with-validation).include?(protocol) ? "https" : "http"
-      URI::Generic.build(:scheme => scheme, :host => server, :port => port, :path => "/nuage/api/#{version}").to_s
+      URI::Generic.build(:scheme => scheme, :host => server, :port => port).to_s
+    end
+
+    def auth_url(protocol, server, port, version)
+      URI(base_url(protocol, server, port)).tap { |url| url.path = "/nuage/api/#{version}" }.to_s
     end
 
     def translate_exception(err)
--- a/app/models/manageiq/providers/nuage/network_manager/floating_ip.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/floating_ip.rb
@@ -1,2 +1,35 @@
 class ManageIQ::Providers::Nuage::NetworkManager::FloatingIp < ::FloatingIp
+  supports :delete
+
+  def delete_floating_ip_queue(userid)
+    task_opts = {
+      :action => "deleting Floating Ip for user #{userid}",
+      :userid => userid
+    }
+    queue_opts = {
+      :class_name  => self.class.name,
+      :method_name => 'raw_delete_floating_ip',
+      :instance_id => id,
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => 'ems_operations',
+      :zone        => ext_management_system.my_zone,
+      :args        => []
+    }
+    MiqTask.generic_action_with_callback(task_opts, queue_opts)
+  end
+
+  def raw_delete_floating_ip
+    $nuage_log.info("Deleting Floating Ip (ems_ref = #{ems_ref})")
+
+    Ansible::Runner.run(
+      ext_management_system.ansible_env_vars,
+      ext_management_system.ansible_extra_vars(:id => ems_ref),
+      ext_management_system.playbook('remove-floating-ip.yml')
+    )
+
+    $nuage_log.info("Done deleting Floating Ip (ems_ref = #{ems_ref})")
+  rescue => e
+    $nuage_log.error "Error deleting Floating Ip: #{e}"
+    raise MiqException::MiqCloudSubnetDeleteError
+  end
 end
--- a/app/models/manageiq/providers/nuage/network_manager/security_group.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/security_group.rb
@@ -1,2 +1,35 @@
 class ManageIQ::Providers::Nuage::NetworkManager::SecurityGroup < ::SecurityGroup
+  supports :delete
+
+  def delete_security_group_queue(userid)
+    task_opts = {
+      :action => "deleting Security Group for user #{userid}",
+      :userid => userid
+    }
+    queue_opts = {
+      :class_name  => self.class.name,
+      :method_name => 'raw_delete_security_group',
+      :instance_id => id,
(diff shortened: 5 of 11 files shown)


app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -1,2 +1,60 @@
...
+    $nuage_log.info("Done deleting Cloud Subnet (ems_ref = #{ems_ref})")
+  rescue => e
+    $nuage_log.error "Error deleting Cloud Subnet: #{e}"
+    raise MiqException::MiqCloudSubnetDeleteError
+  end
+
+  def self.raw_create_cloud_subnet(ext_management_system, options)
+    $nuage_log.info("Create Cloud Subnet (ems_ref = #{ems_ref})")
+
+    # TODO: update UI to match what Nuage needs
```
miha-plesko: @Ladas @agrare I remember you mentioned that UI work is still in progress for AnsibleRunner - does this mean there will be different UI approach for AnsibleRunner inputs? Or should I just go and update the current UI as usually with a few more `ng-if`s to match inputs for creating new Nuage CloudSubnet?

/cc @himdel

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -1,2 +1,60 @@
...
+        :subnet_attributes => { :name => options[:name] }
+      ),
+      ext_management_system.playbook('remove-subnet.yml')
+    )
+
+    $nuage_log.info("Done creating Cloud Subnet (ems_ref = #{ems_ref})")
+
+    # TODO: can playbook return value?
+
+    {:ems_ref => subnet.id, :name => options[:name]}
```
miha-plesko: @Ladas @agrare so here we'd need playbook to return some values. We have two options:

- don't parse playbook output at all
- parse what playbook is printing to the STDOUT

Which approach would you think is better?

app/models/manageiq/providers/nuage/network_manager.rb
```diff
@@ -91,4 +91,53 @@ def l3_cloud_subnets
...
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => 'ems_operations',
+      :zone        => my_zone,
+      :args        => [options]
+    }
+    MiqTask.generic_action_with_callback(task_opts, queue_opts)
+  end
+
+  def create_cloud_subnet(options)
+    CloudSubnet.raw_create_cloud_subnet(self, options)
```
agrare: Looks like this is implicitly relying on this resolving to `ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet` and not `::CloudSubnet`.
Can you be more explicit and do something like `self.class::CloudSubnet`

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -1,2 +1,60 @@
...
+  supports :create
+
+  def delete_cloud_subnet_queue(userid)
+    task_opts = {
+      :action => "deleting Cloud Subnet for user #{userid}",
+      :userid => userid
+    }
+    queue_opts = {
+      :class_name  => self.class.name,
+      :method_name => 'raw_delete_cloud_subnet',
```
agrare: We don't usually (ever?) queue the `raw_*` methods directly, we would queue `delete_cloud_subnet` which would then call `raw_delete_cloud_subnet` directly

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -1,2 +1,60 @@
...
+      :instance_id => id,
+      :priority    => MiqQueue::HIGH_PRIORITY,
```
agrare: Don't think this has to be high priority :)

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -1,2 +1,60 @@
...
+    )
+
+    $nuage_log.info("Done deleting Cloud Subnet (ems_ref = #{ems_ref})")
+  rescue => e
+    $nuage_log.error "Error deleting Cloud Subnet: #{e}"
+    raise MiqException::MiqCloudSubnetDeleteError
+  end
+
+  def self.raw_create_cloud_subnet(ext_management_system, options)
+    $nuage_log.info("Create Cloud Subnet (ems_ref = #{ems_ref})")
```
agrare: Don't think `ems_ref` is going to work here

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -1,2 +1,60 @@
...
+    subnet = Ansible::Runner.run(
+      ext_management_system.ansible_env_vars,
+      ext_management_system.ansible_extra_vars(
+        :domain_id => options[:router_id],
+        :subnet_attributes => { :name => options[:name] }
+      ),
+      ext_management_system.playbook('remove-subnet.yml')
+    )
+
+    $nuage_log.info("Done creating Cloud Subnet (ems_ref = #{ems_ref})")
```
agrare: Same here

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -1,2 +1,60 @@
...
+  end
+
+  def raw_delete_cloud_subnet
+    $nuage_log.info("Deleting Cloud Subnet (ems_ref = #{ems_ref})")
+
+    Ansible::Runner.run(
+      ext_management_system.ansible_env_vars,
+      ext_management_system.ansible_extra_vars(:id => ems_ref),
+      ext_management_system.playbook('remove-subnet.yml')
+    )
```
agrare: I like how this came out

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
agrare: I think returning the ref and name is a good thing in case the caller wants to do a state-machine loop waiting for it to show up in our inventory.  If this is available in the playbook we should print it and parse the stdout

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
agrare: Oh I see, this is how openstack does it.  I'm okay keeping it consistent with how they define it then.

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
agrare: If there is an existing button and action for these methods I don't think you need to make any UI changes

agrare:
@miha-plesko are these roles available on ansible galaxy?  We want to allow people to add roles here but encourage them to be seeded from galaxy.

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
Ladas: I will prepare for you the parsed_stdout, where you should be able to find it.

E.g. in my example, I use ```set_stats``` module to store result:

https://github.com/ManageIQ/manageiq-providers-amazon/pull/453/files#diff-5f61cf6a1eccbf28f5617e02b272fdd1R11

Then doing ```pp parsed_stdout.last["event_data"]["artifact_data"]```, should give all the set_stat records, like this:

```ruby
{"created_security_group"=>
  {"ip_permissions"=>
    [{"from_port"=>22,
      "ip_protocol"=>"tcp",
      "to_port"=>22,
      "ip_ranges"=>[{"cidr_ip"=>"0.0.0.0/0"}],
      "prefix_list_ids"=>[],
      "ipv6_ranges"=>[],
      "user_id_group_pairs"=>[]}],
   "description"=>"ladas_test_changed",
   "tags"=>{},
   "changed"=>false,
   "group_name"=>"ladas_test5",
   "failed"=>false,
   "vpc_id"=>"vpc-ff49ff91",
   "group_id"=>"sg-14ca085e",
   "ip_permissions_egress"=>
    [{"user_id_group_pairs"=>[],
      "prefix_list_ids"=>[],
      "ip_ranges"=>[{"cidr_ip"=>"0.0.0.0/0"}],
      "ip_protocol"=>"-1",
      "ipv6_ranges"=>[]}],
   "owner_id"=>"200278856672"}}
```

Ladas: note: this is returned also if we update the security group, so we should also check if the particular play was changed. Or maybe it doesn't matter that much

app/models/manageiq/providers/nuage/network_manager.rb
```diff
@@ -91,4 +91,53 @@ def l3_cloud_subnets
...
+
+  def ansible_env_vars
+    {}
+  end
+
+  def ansible_extra_vars(extra)
+    {
+      :nuage_auth => {
+        :api_username   => default_authentication.userid,
+        :api_password   => default_authentication.password,
```
Ladas: does the module take these as ENV vars? I think the extravars can be logged, so it's better to pass secrets through env vanrs

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
Ladas: The new way, that the UI team prepares, is that you can have your own javascript code, with the specific form, under the Nuage provider codebase

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
Ladas: also set_stats is cumulative, so calling it multiple times with the same var, should result in having array with all the values

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
Ladas: (I would lean to fix it here, then in OpenStack)

miha-plesko: Yeah, even for OpenStack it's not the same everywhere (delete_cloud_subnet_queue invokes raw directly, while create_cloud_subnet_queue does the additional hop as you suggest). Thanks for guideline, I'll do as you suggest.

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
miha-plesko: Copy-pasted from OpenStack 😇 It sounds reasonable to have user operations high priority because usually she waits to see immediate reaction, pressing F5 all the time, right?

Would you suggest normal priority anyway?

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
miha-plesko: AnsibleRunner is love at first sight for me as well 😍

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
miha-plesko: As I said, AnsibleRunner is a love at first sight 😍 Thanks, what you describe sounds like really cool and managable, can't wait ansible-runner PR being merged.

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
agrare: Yeah that was incorrect when openstack did it as well, we should only use high priority when something _needs_ to jump ahead of everything else in the queue for things to work.

miha-plesko:
@agrare @Ladas roles are not on Ansible Galaxy yet, wanted to come up with something that works first (I think Galaxy support is not there yet, right?). Can you please advise me how to proceed:

- should we wait for the UI changes that @Ladas mentioned or should I ng-if the existing UI?
- should we wait for [`ansible-runner` PR](https://github.com/ManageIQ/manageiq/pull/17688) (@Ladas do you happen to have an estimate?) to get merged or should I pimp up this PR today and we merge it early?
- is there anything I can do to help you proceed with [drb PR](https://github.com/ManageIQ/manageiq-automation_engine/pull/198)? @agrare do you perhaps have a new estimate on how much time would you need, I see some comments there are pretty consuming right?

Thanks in advance, I like how we're breaking ice together here.

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
Ladas: @martinpovolny looks like this will also touch list toolbars, we can't do the provider specific forms there, right?

Ladas:
@miha-plesko I think we will be able to merge the ansible runner today. You might need to tweak the path to executable, until we release the rpm with 1.0.4 (https://github.com/ManageIQ/manageiq/pull/17688/files#diff-259e0b014fb9528a44a195b1ba3025a4R45)

So ultimately, you can just have requirements.yml in the provider code and it will be downloaded/updated on build time https://docs.ansible.com/ansible/latest/reference_appendices/galaxy.html#installing-multiple-roles-from-a-file

For the UI, we will have to find a way that the provider name (Nuage, Openstack, etc.) is not mentioned in the UI, otherwise it will probably not pass the review.

app/models/manageiq/providers/nuage/network_manager.rb
miha-plesko: @Ladas module doesn't recognize ENV vars but fortunately it setts `no_log=True` for sensitive parameters which prevents printing it to the log: https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/network/nuage/nuage_vspk.py#L977.

BTW I'm opening a new PR today with only DELETE methods so that we don't rely on UI at this point.

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
martinpovolny: @miha-plesko : We are indeed working on a way to enable the provider authors such as you to independently extend ManageIQ with Ansible. From the UI perspective it right now means enabling providers to add new button groups to toolbars and specificly toolbars displayed in details screens of their relevant entities. In your case it might be a nuage network. When such a button is pressed there are possibilities what will happen.
 1. an API call will be fired (backed by some Ansible playbook)
 2. a dialog will be opened that will allow the user to enter additional information needed for an API call (backed by some Ansible playbook)

For the 2. there are 2 options:
 a) provider team will implement their own dialog in the form of a React class 
 b) provider team will provide a dynamic dialog definition (dynamic dialogs = service dialogs) and the UI will "play" the dynamic dialog.

In case a) the React class is responsible for executing the API call
In case b) data from the dynamic dialog run will be passed to an API endpoint


Some part of that can be seen with some example here:
https://github.com/ManageIQ/manageiq-ui-classic/pull/4262
and here
https://github.com/ManageIQ/manageiq-ui-classic/pull/4231


This should be a way the provider teams can implement (non-core most likely) functionality (mostly) on their own from their repos.

Having said that I don't think that it's what you should do for network subnet. But if you want to go that way we will try to help you.

martinpovolny: Regarding ng-ifs and provider name: We don't want to get to a spaghetti state (that we already have in places). Specifically I am afraid of provider name tests spread between Ruby and Javascript code. That get's quickly unmaintainable.

Ideally providers should have capabilities and the UI should query those capabilities and behave according to such capabilities and not based on matching a substring in provider name for example.

I perfectly understand that in some cases this is not possible and there needs to be a test for provider name somewhere. I'd just prefer that to be in as few places as possible.

For example if a form needs to change based on provider type (and there are no provider capabilitied that can be tested) you can create a form configuration hash that will drive the indivitual UI elements and have that hash pre-filled bases on provider type on the Ruby side in controller rather than create complex conditions on the javascript side with many ifs and ands. The javascript side then would not contain knowledge about the provider but would instead offer flexibility needed to satisfy different providers' needs.

It works better to talk code with examples rather than talk words in this case so I suggest you ping me early when you start extending existing form so that we can discuss the possibilities w/o you spending time and being afraid that your work would not get merged.

PR #104, opened 2018-07-18
Support DELETE operations powered by AnsibleRunner

miha-plesko:
With this commit we support DELETE operations for following three entities:

- CloudSubnet
- FloatingIp
- SecurityGroup

They are all implemented with AnsibleRunner. Ansible Galaxy role is used to support AnsibleRunner to avoid having role code in MIQ codebase.

Related PR: https://github.com/ManageIQ/manageiq-providers-nuage/pull/103

RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1574972


miha-plesko:
@agrare @Ladas thanks for review on https://github.com/ManageIQ/manageiq-providers-nuage/pull/103. I'm opening a this PR to address your comments as well as provide rspecs. Kindly ask for review.

Did I pick correct `requirements.yml` location for Galaxy? Also, I'm curious how Galaxy roles are going to be managed on appliance. It feels strange that I can in fact modify it at any point (say two days before release 😈 ) and entire CRUD will crash for Nuage. Or I'm missing something?

agrare:
@miha-plesko Ha so re: galaxy roles, we plan to seed the roles from galaxy locally and put them in a features.yml file so that ansible knows not to go out to galaxy at runtime.  It will require a PR to updated them in the provider though.

miha-plesko:
FYI: updated the Galaxy namespace few minutes ago from `miha_plesko.role -> xlab_si.role` now that https://github.com/ansible/galaxy/issues/923 has been resolved.

app/models/manageiq/providers/nuage/network_manager/floating_ip.rb
```diff
@@ -1,2 +1,40 @@
 class ManageIQ::Providers::Nuage::NetworkManager::FloatingIp < ::FloatingIp
+  supports :delete
+
+  def delete_floating_ip_queue(userid)
```
Ladas: we don;t have to have explicit todo here, but we should move this to core too. Only the ```raw_``` method and ```supports <action>```, should be in provider. (for shared actions)

But lets do this in follow up PR

content_tmp/ansible/remove-subnet.yml
```diff
@@ -0,0 +1,14 @@
...
+# Playbook Inputs:
+# - id (string) entity id
+# - kind (string) 'L2' for L2 subnets and 'L3' for L3 subnets (default)
+
+- name: Remove Subnet
+  hosts: localhost
+  connection: local
+  gather_facts: False
+  vars:
+    type: L3
```
Ladas: should this be called ```kind```? I don't see ```type``` anywhere else

miha-plesko: Thanks! Was struggling between `kind` and `type` so much that obviously I mixed it up eventually 😎

content_tmp/ansible/roles/requirements.yml
```diff
@@ -0,0 +1,2 @@
+---
+- src: xlab_si.nuage_remove_entity
```
Ladas: @Fryguy @agrare so, we should prepare the support for this. In which repo/step we should install these? I would expect that bin/update and setup should also update the roles.

@miha-plesko we should probably have version of each role listed explicitely, so we can control it in backports.

Ladas:
@miha-plesko looks like a rebase is needed

So each role should have  ```version``` listed in the requirements.yml, to avoid breakage by updating the role in galaxy.

After that I am ready to merged this. And parallel to that we need to add support for ```requirements.yml``` and add the ```ansible-runner``` binary to the appliance.

miha-plesko:
@Ladas done - rebased, applied comments, retested. I think we're good to go. Is great we merge so we have base ansible functions upstreamed before I start working on UI for CREATE with Martin.

miha-plesko:
@Ladas I've boasted to @Loicavenel that this is already merged only I realised it isn't really. I think you only need to hit the button here, right?

PR #105, opened 2018-07-18
Properly subclass CloudSubnet

miha-plesko:
With this commit we fix current strange subclassing:

```
ManageIQ::Providers::Nuage::NetworkManager::CloudSubnetL2
ManageIQ::Providers::Nuage::NetworkManager::CloudSubnetL3
```

into proper one (notice `::L2` and `::L3` at the end):

```
ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet::L2
ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet::L3
```


miha-plesko:
Implementing AnsibleRunner CRUD I noticed this inconsistence, please let me know whether what I suggest here is cool so that I also quickly update [schema migration PR](https://github.com/ManageIQ/manageiq-schema/pull/215) since it's still not merged.

miha-plesko:
@Ladas are you good with this?

PR #107, opened 2018-07-20
Narrow down focus of targeted refresh

miha-plesko:
When event reports modification of CloudSubnet current targeted refresh performs way too much work: it refreshes not only this specific CloudSubnet, but also all related entities - CloudTenant, NetworkRouter, SecurityGroup etc. This results in 40 API calls instead of one!

Rationale behind was that we wanted to make sure all the relevant entities get updated as well BUT it turns out Nuage yields events for every piece of change anyway so we can safely assume that targeted refresh will handle them.

With this commit we simplify targeted refresh to not infer any related entities and rather just focus on updating the specific one that it was invoked for.


Ladas:
@miha-plesko lets rebase

miha-plesko:
@Ladas rebased. Please let me know whether you prefer to get rid of `infer_related_ems_refs!` entirely or should I leave it there empty?

Ladas:
@miha-plesko we can leave it for now. We will see later if there is some critical path, like with the vm/interfaces/disks example

PR #109, opened 2018-07-20
Targeted refresh for L2 and L3 CloudSubnets

miha-plesko:
With this commit we adjust targeted refresh to works for both kind of subnets - CloudSubnet::L2 and CloudSubnet::L3. When we recieve event we need to pass additional information to target in order to later know what type of subnet are we updating. We need to know this because different API calls need to be used for L2 and L3 subnets...

### DIFF
--- a/app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
@@ -16,13 +16,13 @@ class ManageIQ::Providers::Nuage::Inventory::Collector::TargetCollection < Manag
   end
 
   def cloud_subnets
-    return [] if references(:cloud_subnets).blank?
-    references(:cloud_subnets).collect { |ems_ref| cloud_subnet(ems_ref) }
-    @cloud_subnets_map.values.compact
+    return [] if (refs = references_with_kind(:cloud_subnets, 'L3')).blank?
+    refs.map { |ems_ref| cloud_subnet(ems_ref) }.compact
   end
 
   def l2_cloud_subnets
-    [] # TODO(miha-plesko): targeted refresh for l2_cloud_subnets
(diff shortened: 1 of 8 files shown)


miha-plesko:
@Ladas this one is a bit more interesting because I had to figure how to pass subclass information from event JSON down to the collector. Looking forward to your opinion!

app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
```diff
@@ -16,13 +16,15 @@ def cloud_networks_floating
...
-    return [] if references(:cloud_subnets).blank?
+    return [] if references_with_kind(:cloud_subnets, 'L3').blank?
     references(:cloud_subnets).collect { |ems_ref| cloud_subnet(ems_ref) }
     @cloud_subnets_map.values.compact
   end
 
   def l2_cloud_subnets
-    [] # TODO(miha-plesko): targeted refresh for l2_cloud_subnets
+    return [] if references_with_kind(:cloud_subnets, 'L2').blank?
+    references(:cloud_subnets).collect { |ems_ref| l2_cloud_subnet(ems_ref) }
```
Ladas: so why not use ```references_with_kind(:cloud_subnets, 'L2')``` here?

you can do
```
def l2_cloud_subnets
  return [] if (refs = references_with_kind(:cloud_subnets, 'L2')).blank?
  refs.map { |ems_ref| l2_cloud_subnet(ems_ref) }.compact
end
````

same for the other one

miha-plesko: Thanks for noticing, would cause us headaches in future. I'm not sure why I we don't use `map` with other functions as well instad referencing internal hashes. Will fix in followup PR.

PR #110, opened 2018-07-20
Targeted refresh for FloatingIp

miha-plesko:
What title says ^


miha-plesko:
Thanks @Ladas for noticing we do too much memoization. I've fixed it for FloatingIps in this PR and will prepare a followup PR to remove unneeded memoization all over the collector.

PR #111, opened 2018-07-20
Targeted refresh for NetworkPorts

miha-plesko:
What title says ^


spec/models/manageiq/providers/nuage/network_manager/vcr_specs/targeted_refresh_spec.rb
```diff
@@ -71,7 +78,7 @@
 
         describe "on empty database" do
           it "will refresh cloud_subnet" do
-            cloud_subnet = FactoryGirl.build(:cloud_subnet_nuage, :ems_ref => cloud_subnet_ref1)
+            cloud_subnet = FactoryGirl.build(:cloud_subnet_l3_nuage, :ems_ref => cloud_subnet_ref1)
```
Ladas: so here, :ext_management_system is missing. Has to be filled everywhere, if the table has :ems_id column

miha-plesko:
@Ladas this PR concludes targeted refresh backend. Then we only need the AE Instance in miq-content to make sure it's triggered plus the memoization followup PR as discussed on gitter.

PR #112, opened 2018-07-23
Reduce memoization in targeted refresh

miha-plesko:
We're doing a lot of unneccessary caching in targeted collector. Not anymore - with this commit we simplify the code and spare memory. Rspecs are left intact and there are no errors about missing http interactions which confirms that caching was never used - because each entity is fetched and parsed only once anyway.


bronaghs:
@miha-plesko  Any idea what the memory savings are?

miha-plesko:
@bronaghs haven't measured memory change but IMAO most notable thing here is that code became more readable. Memory change is not too big I think because targeted refresh works with small set of entities anyway.

PR #114, opened 2018-07-24
Infer related CloudSubnet in case of NetworkPort refresh

miha-plesko:
When refreshing NetworkPort we're also updating its many-to-many relation with CloudSubnet (to store address in the intermidiate table). And this fails with ugly integritiy error if subnet is not fetched yet.

With this commit the interity error by making sure subnet is also fetched together with port.

### DIFF
--- a/app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
@@ -157,10 +157,10 @@ class ManageIQ::Providers::Nuage::Inventory::Collector::TargetCollection < Manag
     end
   end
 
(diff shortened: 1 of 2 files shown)


miha-plesko:
Marked WIP becaue it just crashed again, need to debug a bit more.

miha-plesko:
Okay so there is one specific network port that causes issues. Not sure why but maybe Ansible that is periodically creating and deleting it somehow produces invalid state for a moment. No matter how I create/delete vport from UI, it always works for me.

Marking subnet_id as NULL-able to prevent refresher from crashing. I don't think we need to worry much about such invalid state.

@Ladas please take a look

PR #115, opened 2018-07-24
Destroy dependent entities when parent is deleted

miha-plesko:
Zombie children were remaining in VMDB when DELETE targeted refresh was triggered on parent entity because Nuage doesn't emit events for underlying children (which get implicitly deleted).

With this commit we tell Rails to always remove all the children in VMDB as well to avoid the zombie thing.

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
@@ -1,5 +1,10 @@
 class ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet < ::CloudSubnet
+  before_destroy :remove_network_ports, :prepend => true
   supports :delete
+
+  def remove_network_ports
+    network_ports.each(&:destroy)
+  end
 
   def delete_cloud_subnet_queue(userid)
     task_opts = {
--- a/app/models/manageiq/providers/nuage/network_manager/cloud_tenant.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/cloud_tenant.rb
@@ -1,4 +1,10 @@
 class ManageIQ::Providers::Nuage::NetworkManager::CloudTenant < ::CloudTenant
+  has_many   :security_groups, :dependent => :destroy
+  has_many   :cloud_subnets,   :dependent => :destroy
+  has_many   :network_ports,   :dependent => :destroy
+  has_many   :network_routers, :dependent => :destroy
+  has_many   :floating_ips,    :dependent => :destroy
+
   def l3_cloud_subnets
     cloud_subnets.where(:type => self.class.parent.l3_cloud_subnet_type)
   end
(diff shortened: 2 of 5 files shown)


app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -1,2 +1,7 @@
 class ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet < ::CloudSubnet
+  before_destroy :remove_network_ports, prepend: true
+
+  def remove_network_ports
```
miha-plesko: @Ladas so this manual deletion is needed because Rails won't remove actual NetworkPorts based on `:dependent => :destroy` but only the records in intermediate table.

Ladas: hm, we usually delete the network ports as part of the Vms

can you describe a detail usecase/workflow for this to happen?

miha-plesko:
This is what happens when user deletes Enterprise i.e. CloudTenant:

- user deletes CloudTenant on Nuage server
- Nuage server emits event to notify MIQ that CloudTenant was deleted
- but in fact Nuage server removes no only CloudTenant, but also all its children: routers, floating ips, subnets, ports, security groups etc.
- but MIQ only recieved event that CloudTenant was deleted
- by default CloudTenant removal doesn't destroy all children
- so we update logic here to match what happens on Nuage server

It's similar when user removes NetworkRouter on NuageServer: all its children are removed but only one event is emitted.

In other words: we have Ansible running in development environment and it periodically creates some entities and then removes tenant. Result was that VMDB was having more and more CloudSubnets and NetworkPorts with same name. Full EMS refresh removes them, but I would prefer to remove them earlier. Looking forward for your opinion.

app/models/manageiq/providers/nuage/network_manager/cloud_tenant.rb
```diff
@@ -1,4 +1,10 @@
 class ManageIQ::Providers::Nuage::NetworkManager::CloudTenant < ::CloudTenant
+  has_many   :security_groups, :dependent => :destroy
+  has_many   :cloud_subnets,   :dependent => :destroy
+  has_many   :network_ports,   :dependent => :destroy
```
Ladas: so, this should remove all network_ports, is the case that these are some owned by other tenant?

Ladas: ok, makes sense. I don't think we have a better way to do this now

Ladas: btw. nuage doesn't send events for all the objects that are deleted?

miha-plesko: As far as I've tested it doesn't. It only sends event for explicitly deleted entity while it silently removes its children..

Ladas: hm strange, in other providers, we usually get event per each entity

PR #116, opened 2018-07-25
Check for invalid api_version string

miha-plesko:
When user forgets to pick API version upon Nuage provider creation she gets very strange error. Problem is that in such cases our Angular passes in literally value (with questionmarks and all):

```
'? string:v2 ?'
```

This is then used to compose URL and BOOOM. With this commit we therefore manually chech that version looks like expected e.g. `v5_0` and raise error otherwise with descriptive error message which is displayed to user upon credentials validation.

Before 👎 👎 👎 :
image

After 👍 👍 👍 :
image


agrare:
:+1: with backend validation since someone can create a provider e.g. thorough the API.  You could/should probably add a before_save validation as well similar to what is done with the hostname https://github.com/ManageIQ/manageiq/blob/master/app/models/ext_management_system.rb#L80-L81

PR #117, opened 2018-07-25
Implement CREATE operation for CloudSubnet

miha-plesko:
With this commit we provide backend logic for CloudSubnet:L3 creation as well as corresponding playbook. Playbook first creates Zone on Nuage server (if not exist) and then adds desired Subnet to it.

### DIFF
--- a/app/models/manageiq/providers/nuage/ansible_crud_mixin.rb
+++ b/app/models/manageiq/providers/nuage/ansible_crud_mixin.rb
@@ -1,0 +1,19 @@
+module ManageIQ::Providers::Nuage::AnsibleCrudMixin
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def ansible_stats(ansible_response, *keys, default: {})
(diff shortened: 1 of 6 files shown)


miha-plesko:
Kindly ask for your opinion guys, but please don't merge yet as some adjustments might be needed when I do UI for this tomorrow.

Ladas:
@miha-plesko looks like we'll need a rebase

agrare:
So we don't have ansible runner installed on any appliances yet so this should be backed out until that is ready.

miha-plesko:
We're developing React component for it and it's easier if backend is prepared. Are you saying that we should revert commit or just make it dummy function that does nothing?

agrare:
That's up to you guys but just know that this won't actually work yet to anyone without installing ansible-runner manually.

Ladas:
@agrare we have other actions already merged, @miha-plesko is ok with installing it manually until it lands in an appliance (it should be there soon)

PR #122, opened 2018-08-06
Setup JavaScript tests for Nuage provider

miha-plesko:
With this commit we setup two libraries for javascript tests:
    
- **jest** (pojo javascript tests)
- **enzyme** (React component tests)
    
For each of them we provide a single dummy test just to be sure configuration is valid.

Depends on: https://github.com/ManageIQ/manageiq-providers-nuage/pull/121

/cc @Hyperkid123


miha-plesko:
Note: only 2nd commit is important here

package.json
```diff
@@ -15,6 +15,9 @@
   "homepage": "https://github.com/ManageIQ/manageiq#readme",
   "dependencies": {
     "@manageiq/react-ui-components": "~0.9.5",
+    "chai": "^4.1.2",
```
Hyperkid123: Is there any reason why you want to use this library? I don't think you will need this at all. Jest already has pretty good assertions.

miha-plesko: Cool, was looking at enzyme example and they are using it. Let me remove.

spec/javascript/components/components.test.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
+import { Spinner } from 'patternfly-react';
+
+describe('<Spinner/>', () => {
+  it('renders three <Foo /> components', () => {
+    const wrapper = shallow(<Spinner loading />);
+    expect(wrapper.find('.spinner')).to.have.length(1);
```
Hyperkid123: So this is basically one the most annoying things about testing react components. Hopefully you can use snapshot testing with Jest https://jestjs.io/docs/en/snapshot-testing (also i would recommend using [enzime-to-json](https://github.com/adriantoine/enzyme-to-json) instead of the react-renderer in the example).

package.json
```diff
@@ -0,0 +1,52 @@
...
+  },
+  "author": "ManageIQ",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/ManageIQ/manageiq/issues"
+  },
+  "homepage": "https://github.com/ManageIQ/manageiq#readme",
+  "dependencies": {
+    "@manageiq/react-ui-components": "~0.9.5",
+    "enzyme": "^3.3.0",
```
Hyperkid123: `enzyme, enzyme-adapter-react-16, enzyme-to-json` should be dev-dependencies

spec/javascript/components/components.test.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import { Spinner } from 'patternfly-react';
+
+describe('<Spinner/>', () => {
+  it('renders three <Foo /> components', () => {
+    const wrapper = shallow(<Spinner loading />);
+    expect(wrapper.find('.spinner')).toHaveLength(1);
```
Hyperkid123: I would maybe add a snapshot test here to show off how it works:
```
import toJson from 'enzyme-to-json';
...
  expect(toJson(wrapper)).toMatchSnapshot();
...
```

miha-plesko: Yup, I owe you this from yesterday, as soon as Travis configuration works I'll do it.

miha-plesko:
@Hyperkid123 thanks for help, I think we're almost done - Travis is green and we have snapshot test showcase. Please let me know if there is something else we can improve.

Hyperkid123:
I think this should be enough.

PR #126, opened 2018-08-17
Debug log when targeted refresh takes place

miha-plesko:
With this commit we add a simple debug log to easily debug targeted refresh if needed. It outpust following to nuage.log:

```
[----] I, [2018-...] DEBUG -- : target: domain - CREATE - Routy - 143e678c-ef2a-4b36-9bad-a6c157ba2038
[----] I, [2018-...] DEBUG -- : target: policygroup - CREATE - PolicyTEMPLATE - 3a037fd0-41b6-4fbf-a080-06ba7d93daa0
[----] I, [2018-...] DEBUG -- : target: domain - DELETE - Routy - 143e678c-ef2a-4b36-9bad-a6c157ba2038
```


miha-plesko:
(nah, as far as I've tested and seen examples, there was always only a single one. I thing it's a bug to be array, but who am I to judge ;))

PR #127, opened 2018-08-17
[WIP] Overcome another missing event glitch of Nuage server

miha-plesko:
With this commit we overcome a bug on Nuage side that upon deletion of NetworkRouter, corresponding SecurityGroups are also being deleted, but MIQ is not informed about it because no event is emitted.

We overcome this by inferring all related SecurityGroups when "router was deleted" event occurs. Unfortunately, MIQ does not yet modell relation "security group belongs to network router" so we have to perform check for all SecurityGroups in the CloudTenant.

Depends on: https://github.com/ManageIQ/manageiq-providers-nuage/pull/125

Only latest commit is part of this PR, the first one is from #125.

### DIFF
--- a/app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
@@ -173,7 +173,15 @@ class ManageIQ::Providers::Nuage::Inventory::Collector::TargetCollection < Manag
   end
 
   def infer_related_ems_refs_db!
-    # infer related entities based on VMDB here if needed
+    # Delete related security groups upon router deletion.
+    # TODO(miha-plesko): when we have relation "security group belongs to network router" modelled, we can use
+    # dependent: destroy instead inferring here.
+    references_with_options(:network_routers).each do |router|
+      next unless router[:operation] == 'DELETE'
+      if (router = manager.network_routers.find_by(:ems_ref => router[:ems_ref]))
+        router.cloud_tenant.security_groups.each { |group| add_simple_target!(:security_groups, group.ems_ref) }
+      end
+    end
   end
 
   def infer_related_ems_refs_api!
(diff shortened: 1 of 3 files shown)


Ladas:
we can rebase

miha-plesko:
@Ladas rebased, but please let me know if you think this schema change will make it for 4.7 https://github.com/ManageIQ/manageiq-schema/pull/258? If yes, then we can resolve issue better - by leveraging `:dependent => :destroy` (a lot more effective with less API interaction due to infering related targets).

app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
```diff
@@ -173,7 +173,15 @@ def infer_related_ems_refs!
   end
 
   def infer_related_ems_refs_db!
-    # infer related entities based on VMDB here if needed
+    # Delete related security groups upon router deletion.
+    # TODO(miha-plesko): when we have relation "security group belongs to network router" modelled, we can use
+    # dependent: destroy instead inferring here.
+    references_with_options(:network_routers).each do |router|
+      next unless router[:operation] == 'DELETE'
+      if (router = manager.network_routers.find_by(:ems_ref => router[:ems_ref]))
```
Ladas: I think we can do this via 1 query

Something like:

```
manager.security_groups.joins(:cloud_tenant => :network_routers).where({:cloud_tenant => {:network_routers => {:ems_ref => all_router_ems_refs_array}}})
```

miha-plesko: Nice! Let me apply, but since it looks like the schema PR will be taken in, I hope we can omit this query completely. Because the expensive part is that then upon targeted refresh we will be sending one API request per SecurityGroup for all SecurityGroups in entire CloudTenant - while in fact we know upfront that only those related to this network router should be removed. Zero API calls needed, just removing from VMDB 🎆 🍾 😄

miha-plesko:
Closing as it has been resolved via `:dependent =>:destory` in this PR https://github.com/ManageIQ/manageiq-providers-nuage/pull/136

PR #129, opened 2018-08-20
Lock babel because it's in pre-alpha and it's changing

miha-plesko:
With this commit we lock babel to specific version because it's active development causes Travis failures.

@himdel kindly ask for quick check on this. Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "redux": "^4.0.0"
   },
   "devDependencies": {
+    "@babel/runtime": "7.0.0-beta.42",
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-syntax-dynamic-import": "^6.18.0",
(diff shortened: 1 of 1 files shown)


miha-plesko:
For the record, we're copying ui-classic repo with this, see https://github.com/ManageIQ/manageiq-ui-classic/pull/4495

himdel:
LGTM :+1: 

I wish this wasn't necessary, but hopefully we can remove it once a new fixed version comes out.

tadeboro:
I think this problem resolved itself by now, since tests pass on other pull requests that still lack this patch.

himdel:
Interesting, maybe the ui-classic PR helped here after all...

miha-plesko:
Thanks for noticing @tadeboro , it could be that ui-classic PR somehow resolved it for us as well via https://github.com/ManageIQ/manageiq-ui-classic/pull/4495. Closing this one.

PR #130, opened 2018-08-21
Add missing log file before running tests

tadeboro:
Core repository no longer contains log folder, which needs to be
created manually.

This commit simply ensures that log folder is present when checking
out manageiq core repository into spec folder.


tadeboro:
This intentionally only fixes problem with missing `log` folder. Other unrelated test failures are left intact.

miha-plesko:
Thanks @tadeboro , just for the  record: this is the PR that drops `log` directory from git https://github.com/ManageIQ/manageiq/pull/17663

PR #132, opened 2018-08-21
Follow `log` directory creation pattern from other providers

miha-plesko:
With this commit we further improve `log` directory creation now that it's missing from git. This approach is better than what we had before because

a) it uses absolute path and will work regardless $PWD
b) it's executed everytime bin/setup is run, not just upon
   spec/manageiq link creation

Copied from: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/273

/cc @cben @tadeboro

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -6,8 +6,10 @@ gem_root = Pathname.new(__dir__).join("..")
 unless gem_root.join("spec/manageiq").exist?
   puts "== Cloning manageiq sample app =="
   system "git clone https://github.com/ManageIQ/manageiq.git --branch master --depth 1 spec/manageiq"
-  Dir.mkdir("spec/manageiq/log")
 end
+
+# Ensure a log dir always exists.
+gem_root.join("spec/manageiq/log").mkpath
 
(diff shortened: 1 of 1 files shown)


tadeboro:
I feel that this solution does not solve anything. If `spec/manageiq` folder already exists, it means that `bin/setup` already cloned it and ensured that log folder is present (this is what current solution does).

If developer manages `spec/manageiq` contents manually, then I would argue that creating a log folder automatically makes little sense (and may even annoy grumpy developers like me ;).

And about that relative/absolute path: git clone is already done using relative path, so no harm in using relative path there.

cben:
good point (though bin/setup causes stuff to be written to files under `spec/manageiq/log/` anyway).
I don't really care which way we do it, waiting for gitter discussion to settle...

tadeboro:
The stuff that is placed in into `spec/manageiq/log` folder is created by code in the core repo. But yeah, I am also sitting with a popcorn on my lap and waiting for things to settle down;)

miha-plesko:
Closing as issue has been resolved in core repo via https://github.com/ManageIQ/manageiq/pull/17886 so providers no longer need to bother with it.

PR #134, opened 2018-08-22
Targeted refresh: remove subnet when zone is removed

miha-plesko:
There is a known Nuage glitch that some events are not being emitted. With this commit we tackle the one that when zone or zone template are deleted, also all the subnets in it are deleted, but MIQ has no idea about it. So we need to play smart: when recieveing either of

```
nuage_zonetemplate_delete
nuage_zone_delete
```

events, we manually mark also all corresponding subnets for deletion.


miha-plesko:
@Ladas I think this is the last one to fix "missing events glitch". Once we get it in I'll test everything again and hopefully targeted refresh will be fully functional. Well, aside the schema PR that we'll probably have to wait quite a bit to get in.

miha-plesko:
@Ladas you good with this one?

PR #136, opened 2018-08-27
Connect SecurityGroup to NetworkRouter/CloudSubnet

miha-plesko:
With this commit we perform two things:

- inventoring now fetches relation to parent of the SecurityGroup (can   be either NetworkRouter (L3 domain) or CloudSubnet::L2 (L2 Domain))
- set `:destroy => :dependent` because when parent is deleted, SecurityGroup is always deleted as well


miha-plesko:
@Ladas green not that dependent PR has been merged. When we merge this one the topology view will become as fancy as

image

😎

PR #143, opened 2018-09-05
Improve coverage: event catcher and custom toolbar

miha-plesko:
With this commit we provide some additional unit tests to make sure as much code gets executed upon unit testing as possible. Namely:

- tests for event catcher (mocking the qpid gem)
- tests for our custom center-menu button


miha-plesko:
Hey @Ladas I noticed we have coverage below 82% which is the required coverage for supported providers. With this PR I improved the least covered part, the event catcher.

PR #144, opened 2018-09-07
Cleanup refresher code

miha-plesko:
With this commit we cleanup refresher code. There are two pieces of it:

- targeted refresh collector: remove a bunch of unused functions
- full refresh collector: reduce memoization where not needed

All the unit tests still stand.

### DIFF
--- a/app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
@@ -117,52 +117,12 @@ class ManageIQ::Providers::Nuage::Inventory::Collector::TargetCollection < Manag
     safe_list { vsd_client.get_subnets_for_domain(router_ref) }
   end
 
-  def routers_for_tenant(tenant_ems_ref)
-    safe_list { vsd_client.get_domains_for_enterprise(tenant_ems_ref) }
-  end
-
-  def cloud_subnets_for_tenant(tenant_ems_ref)
-    routers_for_tenant(tenant_ems_ref).each_with_object([]) do |d, arr|
-      arr.push(safe_list { vsd_client.get_subnets_for_domain(d['ID']) })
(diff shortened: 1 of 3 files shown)


miha-plesko:
@Ladas while implementing targeted refresh I realized some leftovers are in code that I'd rather remove them, hence the PR.

app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
```diff
@@ -117,52 +117,12 @@ def cloud_subnets_for_router(router_ref)
...
-    return nil if security_group.nil?
-    network_router(security_group['parentID'])['parentID']
-  end
-
   def references(collection)
     target.manager_refs_by_association.try(:[], collection).try(:[], :ems_ref).try(:to_a) || []
   end
 
   def parse_targets!
-    target.targets.each do |t|
```
miha-plesko: This chunk was never used anyway AFAIK since there is currently no option on UI to manually trigger targeted refresh e.g. for CloudSubnet.

app/models/manageiq/providers/nuage/inventory/collector/target_collection.rb
```diff
@@ -210,15 +170,6 @@ def infer_related_ems_refs_api!
         add_simple_target!(:cloud_subnets, subnet['ID'], :options => { :kind => 'L3' })
       end
     end
-
-    references(:network_ports).each do |port|
-      case port['parentType']
-      when 'subnet'
-        add_simple_target!(:cloud_subnets, port['parentID'], :options => { :kind => 'L3' })
```
miha-plesko: This was also wrong and obviously never used as `port` is in fact a string (ems_ref of a network port) so we would crash when accessing `port['parentID']`. Is not needed anyway now that we handle all those "nuage glitches" if you remember.

PR #147, opened 2018-09-13
[GAPRINDASHVILI] Prevent crash upon inventoring SecurityGroups

miha-plesko:
With this commit we fix bug that caused inventoring crash if Nuage server contains at least one SecurityGroup that is connected to L2 domain (as opposed to L3 domains, as implementation assumes).

Problem was that parentID on security group can refer either to L2 domain or L3 domain, but we assumed L3 domain.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1626219


agrare:
@miha-plesko was this a backport of a PR to master?

miha-plesko:
@simaishi green and approved.

PR #154, opened 2018-10-26
Raise error when built-in CRUD playbook fails

miha-plesko:
With this commit we make sure every Ansible::Runner command is followed by "ansible_raise_for_status" check that raises error when playbook execution errored.

Also, we add playbook logging regardless its execution success/failure, but on debug level. This way we will be able to see playbook outputs in log/nuage.log, when desired.

Screenshot of

```
less -r log/nuage.log
```

image

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1642472

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
@@ -1,7 +1,7 @@
 class ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet < ::CloudSubnet
   has_many :security_groups, :dependent => :destroy
 
-  include AnsibleCrudMixin
+  include ManageIQ::Providers::Nuage::AnsibleCrudMixin
 
   before_destroy :remove_network_ports, :prepend => true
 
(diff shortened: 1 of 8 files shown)


Ladas:
@miha-plesko looks like a valid CI failures

miha-plesko:
I'm glad Travis failed, there we were two missing mixin includes that I didn't notice. When you try to have it done quickly before lunch... :)

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -1,7 +1,7 @@
 class ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet < ::CloudSubnet
   has_many :security_groups, :dependent => :destroy
 
-  include AnsibleCrudMixin
+  include ManageIQ::Providers::Nuage::AnsibleCrudMixin
```
miha-plesko: @Ladas any idea why the shorthand include works on CloudSubnet, but not on FloatingIp/SecurityGroup? There it attempts to include it with full namespace like 

```
ManageIQ::Providers::Nuage::NetworkManager::FloatingIp::AnsibleCrudMixin
```

but here on CloudSubnet it just knows the correct namespace 🤔

Ladas: Think the magic loader tries several variants, using the current scope should be first one (so ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet)

So for me it's always safer to just specify the full path

simaishi:
Hammer backport details:

```
$ git log -1
commit 508ec9460a138177283285f6513303b7961a5f94
Author: Ladislav Smola <lsmola@redhat.com>
Date:   Fri Oct 26 14:33:53 2018 +0200

    Merge pull request #154 from miha-plesko/log-playbook-output-for-debug
    
    Raise error when built-in CRUD playbook fails
    
    (cherry picked from commit 4f6c0cc89168df4400e45b8fbb60d7c390165275)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1642472
```

PR #166, opened 2019-03-13
[WIP] Trigger Travis after https://github.com/ManageIQ/manageiq/pull/18544

miha-plesko:
Just triggering Travis to double check if it's 🍏 after shifting qpid_proton in core branch, see https://github.com/ManageIQ/manageiq/pull/18544


miha-plesko:
Travis is green as expected, closing.

PR #173, opened 2019-05-16
[HAMMER] Bump hammer travis.yml to ruby 2.4.6

jrafanie:
Bump hammer travis.yml to ruby 2.4.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-- 2.4.5
+- 2.4.6
 sudo: required
 cache:
   bundler: true


simaishi:
@jrafanie the `exclude` version stays as is?
https://github.com/ManageIQ/manageiq-providers-nuage/blob/cd688aa07de631b71afc3b641cdf05a10a4eb255/.travis.yml#L19

miha-plesko:
@jrafanie we put the exclude there back then when we were rspec testing multiple ruby versions but we wanted to run javascript tests only once. It can safely be removed now, although it doesn't change anything if it stays.

PR #176, opened 2019-07-01
Fix spec failures for ems counts

d-m-u:
So these counts needed to be adjusted to reflect that ems leaf classes have the inheritance set when you use the factory. 

🤷‍♀ 

I can't even actually run specs here at all, because of this:
An error occurred while loading spec_helper.
Failure/Error: require 'qpid_proton'

LoadError:
  cannot load such file -- qpid_proton

but it's highly likely that's just me.

### DIFF
--- a/spec/models/manageiq/providers/nuage/network_manager/vcr_specs/refresher_spec.rb
+++ b/spec/models/manageiq/providers/nuage/network_manager/vcr_specs/refresher_spec.rb
@@ -105,7 +105,7 @@ describe ManageIQ::Providers::Nuage::NetworkManager::Refresher do
   end
 
   def assert_table_counts
-    expect(ExtManagementSystem.count).to eq(1 + 1) # nuage + vmware infra
+    expect(ExtManagementSystem.count).to eq(3) # nuage + vmware infra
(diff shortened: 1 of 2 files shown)


d-m-u:
@miq-bot assign @agrare

d-m-u:
@bdunne got an idea about how to test this in a way that won't fail the unique guid check?

d-m-u:
@agrare I know it's red, but it's red still for reasons that aren't associated with networks or emses. Can you merge this so I can address the other ones in a separate PR please?

agrare:
@d-m-u I think the issue is that a nuage network manager doesn't actually have a parent manager

agrare:
Okay well this fixes some of the issues

agrare:
Oh no duh, right in that spec file there's this gem
`allow(SecureRandom).to receive(:uuid).and_return('11111111-2222-3333-4444-555555555555')`
so of course all the EMSs are going to have the same guid

miha-plesko:
Ouch, yeah so @agrare @bdunne the reason we had to mock random generator was to test event parser which calculates event's EMS ref. The ugly side effect never popped up so far..

PR #177, opened 2019-07-03
Fix specs failing cause we're using the same guid :rage:

d-m-u:
🤷‍♀ 
Sorry Adam but I think this one's gotta go since I don't even know why you would...


d-m-u:
@miq-bot assign @agrare 
It passed here: https://travis-ci.org/d-m-u/manageiq-providers-nuage/jobs/553771877
@simaishi this should fix this repo for all the things I broke anyway till now 😆

miha-plesko:
@d-m-u yeah not to best way to test it but at least I can answer the "why" part - is because event parser was behaving strange when ems_ref isn't set for event and we don't always get usable value to grab as ems_ref; so we decided to use random.

The test could stay if we put the SecureRandom mocking thing into the test itself, but I guess we can let if go as well

PR #179, opened 2019-07-17
Implement relations that are typically delegated to cloud manager

miha-plesko:
Nuage provider is a bit special because it's a network provider only without cloud provider. This brings some problems with delegates that are specified for network manager by default (like .hosts) because they always get resolved as `nil` instead as empty queryset. Some parts of code don't check for nil, like metrics collector:

https://github.com/ManageIQ/manageiq/blob/master/app/models/metric/targets.rb#L98

```
ems.hosts.each { ... } # BOOOM because hosts is nil
```

so the code explodes. With this commit we "fake" all delegates so that they now return empty querysets.

```
ems.hosts.each { ... } # skips because hosts is empty queryset
```

/cc @pdellaert @gberginc

### DIFF
--- a/app/models/manageiq/providers/nuage/cloud_delegates_mixin.rb
+++ b/app/models/manageiq/providers/nuage/cloud_delegates_mixin.rb
@@ -1,0 +1,58 @@
+module ManageIQ::Providers::Nuage::CloudDelegatesMixin
+  def flavors
+    Flavor.none
+  end
+
+  def cloud_resource_quotas
+    CloudResourceQuota.none
+  end
+
+  def cloud_volumes
+    CloudVolume.none
+  end
+
+  def cloud_volume_snapshots
+    CloudVolumeSnapshot.none
+  end
+
+  def cloud_object_store_containers
+    CloudObjectStoreContainer.none
+  end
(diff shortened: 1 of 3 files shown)


miha-plesko:
@agrare there was a report that metrics collection for OpenStack explodes as soon as Nuage provider is added to the same appliance. Then as soon as Nuage is removed, the metrics collection works again. The problem turns to be this line here https://github.com/ManageIQ/manageiq/blob/master/app/models/metric/targets.rb#L98 where we loop thru all manager and call `ems.hosts` on each of them... With Nuage this obviously doesn't work because it has no parent cloud provider.

Now I'm not sure whether we should fix the metrics collection code, but just in case similar loop is done somewhere else I decided to update Nuage instead. Wdyt?

agrare:
Hm I wonder why it doesn't just do `zone.ems_infras` ... let me dig a little bit.  I'd like to fix the core method but in case there was some reason why it tries to collect infra metrics from all emss I'll merge this
Thanks @miha-plesko !

Fryguy:
Closing in favor of ManageIQ/manageiq#19004

PR #181, opened 2019-07-26
[IVANCHUK] Change travis.yml to use the only supported ruby, 2.5.3

d-m-u:
Change travis.yml to use the only supported ruby, 2.5.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
+---
 dist: xenial
 language: ruby
 rvm:
-- 2.4.6
 - 2.5.3
 sudo: required
 cache:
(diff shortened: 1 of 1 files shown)


simaishi:
Just like manageiq-ui-classic... this will cause `yarn:test` to not run at all... can we remove `matrix` section?

FYI, ui-classic is modifying a test to make sure we run test for 1 ruby version: https://github.com/ManageIQ/manageiq-ui-classic/pull/5885

simaishi:
^ cc @miha-plesko

miha-plesko:
Thanks for ping, lemme fix this on Monday

miha-plesko:
@simaishi TBH I don't see how `excludes` block could cause Travis from skipping javascript tests - I think in worst case it'd cause redundant re-runs of it (when there is no match). However, I've introduced rspec to be sure, thanks for pointing out:

- https://github.com/ManageIQ/manageiq-providers-nuage/pull/182 (rspec only)
- https://github.com/ManageIQ/manageiq-providers-nuage/pull/183 (updated travis + rspec)

PR #182, opened 2019-07-29
[IVANCHUK] Unittest .travis.yml to ensure excludes match the matrix

miha-plesko:
When we're shifting supported ruby versions we often forget to update the "excludes" block which in turn means we're then re-running javascript tests more than we need to. Now, there is no big harm in re-running the test more than once, but still.

With this commit we unit test that excludes block matches values from the matrix. The rspec is actually copy-pasted from ui-classic repo (thanks @himdel and @mzazrivec )

### DIFF
--- a/spec/other/travis_spec.rb
+++ b/spec/other/travis_spec.rb
@@ -1,0 +1,19 @@
+describe 'travis.yml' do
+  let(:travis) { YAML.safe_load(File.open(ManageIQ::Providers::Nuage::Engine.root.join('.travis.yml'))) }
+  let(:versions) { travis['rvm'] }
+
+  it "matches versions and excludes for multiple ruby versions" do
+    if versions.length > 1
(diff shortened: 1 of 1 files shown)


himdel:
2 concerns:

* shouldn't this be a master PR with ivanchuk/yes ?
* https://github.com/ManageIQ/manageiq-ui-classic/pull/5885/files#r308168006

miha-plesko:
@himdel I've addressed both comments, thanks. I tried to speedup merge, hence different PR for ivanchuk, bacause for master we first need to fix .travis.yml, only then we can provide the rspec; while for ivanchuk the .travis.yml is okay already. Anyway, here's the solution:

- https://github.com/ManageIQ/manageiq-providers-nuage/pull/184#issuecomment-515965702 (master only, to fix travis.yml)
- https://github.com/ManageIQ/manageiq-providers-nuage/pull/183 (master & ivanchuk, but will be red on master until the 1st PR is merged)

PR #201, opened 2020-03-27
Move qpid_install to standard location

agrare:


tools/ci/before_install.sh
```diff
@@ -20,3 +21,6 @@ make all
 
 # Install system libraries
 sudo make install
+
+cd $TRAVIS_BUILD_DIR
```
agrare: Interestingly without this the cwd is left as `$HOME/build/qpid-proton/build` (not sure why not just `$HOME/qpid-proton/build` but whatever) and the next `bin/setup` command fails

Fryguy: oh that makes sense because of all of the other commands above

```
cd $HOME/build
...
cd qpid-proton
...
cd build
```

We should probably pushd/popd in this script, but that can be a future thing.

agrare: Wow I've never heard of those, makes sense doing that now

PR #203, opened 2020-04-08
Update the qpid-proton C library version to 0.30.0

agrare:
We bumped the qpid_proton gem version to 0.30.0 in https://github.com/ManageIQ/manageiq/pull/20016, update the C library to
the same version.

PR #204, opened 2020-04-08
Use pushd/popd instead of cd at the end of the script

agrare:
Follow up to: https://github.com/ManageIQ/manageiq-providers-nuage/pull/201#discussion_r405742793

Just needs a rebase after https://github.com/ManageIQ/manageiq-providers-nuage/pull/203

### DIFF
--- a/tools/ci/before_install.sh
+++ b/tools/ci/before_install.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -v
+pushd $PWD
 
 # Install the dev dependencies for building Qpid proton system library.
 sudo apt-get install -y gcc cmake cmake-curses-gui uuid-dev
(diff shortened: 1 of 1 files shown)


tools/ci/before_install.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -v
+pushd $TRAVIS_BUILD_DIR
```
Fryguy, line 3: Would it be better to do `pushd $(pwd)`?

Fryguy: or even `pushd .`

agrare: actually I can just do `pushd $PWD` its already in a var

PR #206, opened 2020-04-09
Use :payload instead of :data, set :proxy, in Rest class

djberg96:
Within the `ManageIQ::Providers::Nuage::NetworkManager::VsdClient::Rest#request`, which is a wrapper for `RestClient::Request#execute`, any requests with a payload will currently fail.

The problem is that the current code is setting the `:data` option, when in fact rest-client uses a `:payload` option:

https://github.com/rest-client/rest-client/blob/master/lib/restclient/request.rb#L54
https://github.com/rest-client/rest-client/blob/master/lib/restclient/request.rb#L163

Based on my (admittedly vague) recollection from my days writing azure-armrest, even a nil `:payload` will cause problems for requests that don't support them, so that key is only added to the hash of options if the payload is non-nil.

While I was here I also added the `:proxy` option.

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/vsd_client/rest.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/vsd_client/rest.rb
@@ -58,14 +58,21 @@ class ManageIQ::Providers::Nuage::NetworkManager::VsdClient::Rest
   def request(url, method: :get, data: nil, verify_ssl: false)
     $nuage_log.debug("Accessing Nuage VSD url #{method} #{url} with data '#{data}'")
     login unless @api_key
-    RestClient::Request.execute(
+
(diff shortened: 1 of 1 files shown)


djberg96:
@agrare should we backport?

simaishi:
Jansa backport details:

```
$ git log -1
commit b6520bb8458636d4759674fec5f5c48af8c9012c
Author: Adam Grare <agrare@redhat.com>
Date:   Thu Apr 9 11:33:59 2020 -0400

    Merge pull request #206 from djberg96/rest_request

    Use :payload instead of :data, set :proxy, in Rest class
    (cherry picked from commit eb7435a062ec1306179b99575b0fed410b90a7ec)
```

PR #210, opened 2020-04-30
Convert the Cloud Subnet Form to DDF and get rid of react-ui-components

skateman:
This form is accessible from a Nuage Network Router's summary screen:
Screenshot from 2020-04-30 13-42-24

The look & feel should be the same, a lot of things got simplified, I used the FirmwareRegistryForm from the ui-classic as an inspiration. There's a piece of error handling that I changed. If the API returns an error while fetching the network router's data, I am displaying it as a flash message instead of printing it on the modal. The reason for this is because there's already a modal when an API error occurs.

**Before:**
Screenshot from 2020-04-30 13-53-32

**After:**
Screenshot from 2020-04-30 13-46-40

Except of this minor change, everything should behave the same.

This change allowed me to get rid of the `react-ui-components` dependency in this repo, which resolves one more checkbox in https://github.com/ManageIQ/manageiq-ui-classic/issues/6716#issuecomment-591974680

@miq-bot add_reviewer @himdel 
@miq-bot add_label react, cleanup


skateman:
I guess travis will fail until https://github.com/ManageIQ/manageiq-ui-classic/pull/7005 is not merged

app/javascript/components/create-nuage-cloud-subnet-form.jsx
```diff
@@ -1,50 +1,59 @@
...
-        loading={this.state.loading}
-      />
-    );
-  }
-}
+  return (
+    <MiqFormRenderer
+      schema={createSchema()}
+      onSubmit={submitValues}
+      buttonsLabels={{ submitLabel: __('Create') }}
```
skateman, line 51: This is not necessary :scissors:

PR #229, opened 2020-10-26
Fix .rubocop_local.yml

Fryguy:
Fix .rubocop_local.yml


bdunne (commented):
Can you fix the newline at the end too?

PR #231, opened 2020-10-27
[JANSA] Test with ruby 2.6.6 on Jansa

jrafanie:
Test with ruby 2.6.6 on Jansa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,7 @@
 dist: xenial
 language: ruby
 rvm:
-- 2.5.7
-- 2.6.5
+- 2.6.6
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,7 +18,7 @@ env:
   - TEST_SUITE=yarn:test
 matrix:
   exclude:
-  - rvm: 2.5.7
+  - rvm: 2.6.6
```
jrafanie, line 21: one second... I need to fix this

jrafanie:
Failure with blaster outer... starting over.

PR #249, opened 2021-06-28
Refactor delete subnet queue

nasark:
- `delete_cloud_subnet_queue` was removed and added to core

Depends on:

* [x] [Refactor delete subnet queue from providers](https://github.com/ManageIQ/manageiq/pull/21303)

### DIFF
--- a/app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
@@ -40,23 +40,6 @@ class ManageIQ::Providers::Nuage::NetworkManager::CloudSubnet < ::CloudSubnet
     network_ports.each(&:destroy)
   end
 
-  def delete_cloud_subnet_queue(userid)
-    task_opts = {
-      :action => "deleting Cloud Subnet for user #{userid}",
-      :userid => userid
-    }
-    queue_opts = {
-      :class_name  => self.class.name,
-      :method_name => 'delete_cloud_subnet',
-      :instance_id => id,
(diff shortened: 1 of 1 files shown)


nasark:
@miq-bot add-labels refactoring

app/models/manageiq/providers/nuage/network_manager/cloud_subnet.rb
```diff
@@ -40,23 +40,6 @@ def remove_network_ports
...
   end
 
-  def delete_cloud_subnet_queue(userid)
-    task_opts = {
-      :action => "deleting Cloud Subnet for user #{userid}",
-      :userid => userid
-    }
-    queue_opts = {
-      :class_name  => self.class.name,
-      :method_name => 'delete_cloud_subnet',
```
agrare, line 50: The spec is failing because it is looking for a queue item with `:method_name => delete_cloud_subnet` but openstack/vpc were queueing `raw_delete_cloud_subnet`.

The way nuage is doing it here is preferred, sorry I missed the difference in reviewing your core PR.  Could you push a follow-up to https://github.com/ManageIQ/manageiq/pull/21303 to change the queued method to be `delete_cloud_subnet` ?

nasark: Thanks for the catch, sorry I should have noticed that in the Travis build failure

nasark: https://github.com/ManageIQ/manageiq/pull/21307

agrare (approved):

PR #267, opened 2022-02-05
Restore deleted before_install

Fryguy:


Fryguy:
@agrare I'm a little confused why this one has BUNDLE_WITH: qpid_proton, but the openstack provider doesn't in https://github.com/ManageIQ/manageiq-providers-openstack/pull/775

agrare:
:shrug: they were added at totally different times and both were valid ways to do it until now?

PR #299, opened 2023-10-27
Require the gem defining the constant we're subclassing

jrafanie:
* We're not setup to autoload the library so we need to manually require the gem before using it.
* Some environments don't have qpid_proton installed so explicitly tell zeitwerk to not autoload or eager load the messaging handler
* Mark specs requiring qpid_proton and exclude them if not available

Related change that should be merged together: https://github.com/ManageIQ/manageiq/pull/22763

### DIFF
--- a/lib/manageiq/providers/nuage/engine.rb
+++ b/lib/manageiq/providers/nuage/engine.rb
@@ -11,6 +11,12 @@ module ManageIQ
           app.config.paths["config/secrets"] << root.join("config", "secrets.yml").to_s
         end
 
+        initializer 'manageiq-providers-nuage.configure_autoloader' do
+          # The following file should not be eager loaded as it depends on being installed, which isn't possible on macs and isn't done by default
+          # in ci for other plugins, core, etc.
+          Rails.autoloaders.main.ignore(root.join("app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb")) unless Gem.loaded_specs.key?('qpid_proton')
+        end
+
         def self.vmdb_plugin?
           true
         end
--- a/spec/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler_spec.rb
+++ b/spec/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler_spec.rb
@@ -1,4 +1,6 @@
-describe ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler do
+# Note, we're using string class name to avoid autoloading the class since it requires
+# qpid_proton installed.
+describe "ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler", :qpid_proton => true do
   before do
     subject.instance_variable_set(:@conn, conn)
   end
@@ -7,7 +9,7 @@ describe ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHand
   let(:container) { double('container', :stopped => false, :stop => nil) }
   let(:conn)      { double('connection', :open_receiver => nil, :container => container) }
 
-  subject { described_class.new(options) }
+  subject { ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler.new(options) }
 
   it 'default connection timeout' do
     expect(subject.instance_variable_get(:@timeout)).to eq(5.seconds)
--- a/spec/models/manageiq/providers/nuage/network_manager_spec.rb
+++ b/spec/models/manageiq/providers/nuage/network_manager_spec.rb
@@ -72,9 +72,10 @@ describe ManageIQ::Providers::Nuage::NetworkManager do
       end.to raise_error(MiqException::MiqInvalidCredentialsError)
     end
 
-    context 'AMQP connection' do
+    context 'AMQP connection', :qpid_proton => true do
       before do
         @conn = double('connection', :handler => handler)
+        require 'qpid_proton'
         allow(Qpid::Proton::Container).to receive(:new).and_return(@conn)
 
         creds = {}
--- a/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
+++ b/app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb
@@ -1,9 +1,8 @@
+require 'qpid_proton'
 class ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler < Qpid::Proton::MessagingHandler
   attr_reader :errors
 
   def initialize(options = {})
-    require 'qpid_proton'
-
     super()
     @options = options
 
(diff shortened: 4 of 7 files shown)


Fryguy:
LGTM, I think 😅  - I still think we should avoid subclassing Qpid::Proton::MessagingHandler at all, if possible

jrafanie:
> LGTM, I think 😅 - I still think we should avoid subclassing Qpid::Proton::MessagingHandler at all, if possible

💯 Absolutely.  If I knew how to set this up to test without it, I would.

agrare:
```
module Qpid::Proton

  # A handler for AMQP messaging events.
  #
  # Subclass the handler and provide the #on_xxx methods with your event-handling code.
```

Subclassing the Qpid MessagingHandler is the correct way to be doing this unfortunately.  We could try to delay the class definition maybe?

jrafanie:
I'm moving the autoloader ignore from https://github.com/ManageIQ/manageiq/pull/22752 here

jrafanie:
Note, this neither prevent us from making the qpid_proton gem installable on other platforms nor bury the usage of qpid's constants into instance methods, such as if we we use composition instead of inheriting from them.  This is a temporary workaround so we can eager load the manageiq application, such as when we generate the message catalogs,

Fryguy (approved):

Fryguy:
@agrare Please take a look

lib/manageiq/providers/nuage/engine.rb
```diff
@@ -11,6 +11,12 @@ class Engine < ::Rails::Engine
           app.config.paths["config/secrets"] << root.join("config", "secrets.yml").to_s
         end
 
+        initializer 'manageiq-providers-nuage.configure_autoloader' do
+          # The following file should not be eager loaded as it depends on being installed, which isn't possible on macs and isn't done by default
+          # in ci for other plugins, core, etc.
+          Rails.autoloaders.main.ignore(root.join("app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb"))
```
jrafanie, line 17: I'm wondering if I should do a `require 'qpid_proton' rescue... ^^^` that line here so we don't need to require the handler in the spec.  Autoload from within the engine should work if qpid_proton is installed.

jrafanie: The problem with this is that all production appliances/pods where qpid_proton is installed will eager require qpid_proton when this engine is initialized.

agrare: In openstack we exclude it from the rspec configuration if qpid_proton isn't loaded could we do that here as well?

https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/spec/spec_helper.rb#L11-L13

jrafanie: yes, that works, thank you

agrare:
@jrafanie what happens if you run `rspec` ?  I'm assuming it will fail on `spec/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler_spec.rb`
Wonder if we should also filter that from the rspec run if the gem isn't installed like openstack does.

jrafanie:
> @jrafanie what happens if you run `rspec` ? I'm assuming it will fail on `spec/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler_spec.rb`
> Wonder if we should also filter that from the rspec run if the gem isn't installed like openstack does.

I assumed you can't run any of them on mac...

https://github.com/jrafanie/manageiq-providers-nuage/blob/c9a6e050da70ff4e06d2cabc60261edff78c483f/spec/spec_helper.rb#L10

agrare:
Okay I can handle that in a follow-up, but I think we should let the rest of the specs that don't depend on qpid_proton still run

jrafanie:
> Okay I can handle that in a follow-up, but I think we should let the rest of the specs that don't depend on qpid_proton still run

one sec... I'll try to get any that can't run here

lib/manageiq/providers/nuage/engine.rb
```diff
@@ -11,6 +11,12 @@ class Engine < ::Rails::Engine
           app.config.paths["config/secrets"] << root.join("config", "secrets.yml").to_s
         end
 
+        initializer 'manageiq-providers-nuage.configure_autoloader' do
+          # The following file should not be eager loaded as it depends on being installed, which isn't possible on macs and isn't done by default
+          # in ci for other plugins, core, etc.
+          Rails.autoloaders.main.ignore(root.join("app/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler.rb")) unless Gem.loaded_specs.key?(:qpid_proton)
```
agrare, line 17: Interesting seems like this should be a string key `Gem.loaded_specs.key?("qpid_proton")`  Wonder when that changed, we should change openstack if it is an issue there as well.
With that it is green for me locally.

spec/models/manageiq/providers/nuage/network_manager/event_catcher/messaging_handler_spec.rb
```diff
@@ -1,4 +1,6 @@
-describe ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler do
+# Note, we're using string class name to avoid autoloading the class since it requires
+# qpid_proton installed.
+describe "ManageIQ::Providers::Nuage::NetworkManager::EventCatcher::MessagingHandler", :qpid_proton => true do
```
agrare, line 3: Actually seems like using strings here is causing the `described_class` below to fail

jrafanie: just pushed a fix to not used `described_class` in the specs

spec/models/manageiq/providers/nuage/network_manager_spec.rb
```diff
@@ -72,7 +72,7 @@
       end.to raise_error(MiqException::MiqInvalidCredentialsError)
     end
 
-    context 'AMQP connection' do
+    context 'AMQP connection', :qpid_proton => true do
```
agrare, line 75: If you throw a
```suggestion
    context 'AMQP connection', :qpid_proton => true do
      require 'qpid_proton'

```

here does that still pass for you without qpid_proton?  Since we're explicitly referencing `Qpid::Proton::Container` below I'm getting
```
Failures:

  1) ManageIQ::Providers::Nuage::NetworkManager validation AMQP connection handles connection errors
     Failure/Error: allow(Qpid::Proton::Container).to receive(:new).and_return(@conn)
     
     NameError:
       uninitialized constant Qpid
```

agrare: The PR is green but I think it is sporadic, if I run this spec file directly and sometimes when running the whole suite it will fail with this error

jrafanie: it doesn't like it out of the before block but works within the before.. pushing

jrafanie: > The PR is green but I think it is sporadic, if I run this spec file directly and sometimes when running the whole suite it will fail with this error

yeah, good point... randomly getting qpid_proton required in other tests BEFORE this line would hide this error

agrare:
Okay I think we're finally there, sorry @jrafanie haha

agrare (approved):

